PR #17, opened 2018-02-22
Add missing string.h includes

lioncash:
Just some inclusions that were missing

### DIFF
--- a/exosphere/gcm.c
+++ b/exosphere/gcm.c
@@ -1,4 +1,5 @@
 #include <stdint.h>
+#include <string.h>
 
 #include "utils.h"
 #include "gcm.h"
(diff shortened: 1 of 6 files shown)


SciresM:
Not sure about this one -- Are we using <string.h> or rolling our own memcpy/memset? Paging @TuxSH since I think he's the person to ask.

lioncash:
What would the benefit of rolling memcpy and memset be? (if it's for perf reasons, have any benchmarks been done to warrant it?)

SciresM:
Per @TuxSH in private conversation: "string.h afaik, but without newlib. We'll see onceit builds"

I think I'll let this sit for now and let Tux figure out how to merge/deal with it.

lioncash:
Sure thing, I've moved the only unrelated header change out into its own commit in #25

SciresM:
Cool, much appreciated. Merged #25. :)

PR #19, opened 2018-02-22
package2: Correct variable name for new device key source

lioncash:
Identifiers can't begin with a number


SciresM:
Any chance you could re-name to "new_devicekey_source_4x"? That's better than numbers-as-words, imo.

lioncash:
Done

SciresM:
Thanks! :)

PR #23, opened 2018-02-23
mmu: Correct TCR_ORGN_WBWA define

lioncash:
Previously this was incorrectly set as the Write-Back-Read-Allocate No Write-Allocate Cacheable define


lioncash:
Ah, beaten to it in #22

PR #26, opened 2018-02-23
mmu: Convert bool to int

lioncash:
Previously this would fail to compile since `bool` (and `true` and `false`) weren't visible


SciresM:
Ahhh stylistically I am actually trying to avoid bool entirely. Better to change the code using it to use int, 1, and 0.

lioncash:
Closing in favor of #30

PR #27, opened 2018-02-23
mmu: Fix compilation issues

lioncash:
mmu_unmap_range was also passing the physical address.

If this is actually unnecessary and the argument itself should be removed instead, then I can just modify this change to do that instead.


SciresM:
This one is, I think, probably @TuxSH's call.

TuxSH:
I think that is unnecessary -- you don't need to know the PA to unmap a given PA.

MerryMage:
Relateldly, there is also a `mmu_unmap_page_range` call (a function that doesn't exist) in `mmio_unmap_all_devices`.

That should likely be:
```
mmu_unmap_range(3, mmu_l3_tbl, MMIO_BASE + offset, devices[i].size);
```
instead.

lioncash:
Amended both

PR #32, opened 2018-02-23
package2: Fix buffer variable name in rsa2048_pss_verify

lioncash:


SciresM:
ahh, this should actually be "message", not "h_buf".

lioncash:
Figured, but erred on the side of caution, since m is so close to h on a keyboard, fixed :)

PR #40, opened 2018-02-23
smc_api: Fix compilation issues

lioncash:
~~The only issue remaining (that GCC has complained about) is line 447 where `devices` can't be found, but I don't know what to replace it with.~~ Nevermind, I misread the error message


MerryMage:
> `devices` can't be found
`#include "mmu.h"`

lioncash:
Thanks!

PR #55, opened 2018-02-26
package2: Separate MAXVER and MINVER maximums from the version numbers

lioncash:
Inevitably, when there's a version update, this will need to be incremented. Therefore,
we make the maximums separate from the version numbers and use a separate define to
indicate that.


SciresM:
The 400_CURRENT naming was actually incredibly intentional. I would like to keep it that way.

lioncash:
OK. Though it should probably be stated somewhere that it is intentional.

PR #65, opened 2018-04-24
fusee: Add tsec fw from 4.x

balika011:


SciresM:
Sorry, can't accept this pull request.

Release atmosphere will search for tsec firmware in NAND, and fall back to reading it from SD.

Issue #66, opened 2018-04-25
Wrong hex value for I2C5 in bootup.c


Resaec:
The hex value of 0x20000000 at line 97 represents I2C4 instead of the mentioned I2C5.

Resaec:
value is correct, comment fixed #67

PR #67, opened 2018-04-25
ENUMs for APB Slave Security Enable registers

Resaec:
I added enums for all APB slave register bits and replaced the magic hex values.

While doing so I noticed that at line 97 in bootup.c the comment says "Also mark I2C5 secure only" while the hex value 0x20000000 coresponds to I2C4.
For the time being I left it at that value and used I2C4 instead of I2C5.

regarding #66


SciresM:
Thanks! :)

Style for enums is typedef enum { } EnumType, with members as ENUMTYPE_MEMBERNAME -- could you refactor?

I'll look into whether what you mentioned is a fuckup in my code or in my comments :)

Resaec:
I'll refactor it in a bit.
Thanks

SciresM:
I checked -- the value ORR'd is 0x20000000. If that's I2C4, my comment was wrong.

PR #71, opened 2018-04-26
Update Read-Me with information on sub projects.

OtherOfIce:
Information pulled from the projects page.


OtherOfIce:
Was just reading through the [Reswitched weekly 15](https://reswitchedweekly.github.io/ReswitchedWeekly15/)  seems like some of the descriptions provided there would be better. Also the post seems to imply that exosphere is done ?

hexkyz:
The current layout of the project provides individual README files for each stage with brief descriptions that may change in the future. Closing this PR until more in-depth explanations/documentation becomes necessary.

PR #77, opened 2018-05-03
Replace std::make_tuple with simpler syntax

leoetlino:
It appears that Stratosphère is targeting C++17. In C++17, std::make_tuple is not required for initialisating a tuple anymore. Same thing, but less typing.


SciresM:
If we're going to switch to this formatting for tuples, could you also fix in all the service funcs that return tuples?

leoetlino:
Sure, I've replaced all the other `std::make_tuple`s with the newer syntax.

Also fixed the missing returns here:

https://github.com/Atmosphere-NX/Atmosphere/blob/4199be2460904d015acdafa85c6b2c805226e15d/stratosphere/loader/source/ldr_process_manager.cpp#L42

https://github.com/Atmosphere-NX/Atmosphere/blob/4199be2460904d015acdafa85c6b2c805226e15d/stratosphere/loader/source/ldr_process_manager.cpp#L47

PR #85, opened 2018-05-05
boot: Fix usage of svcQueryIoMapping and building

leoetlino:
* Fix invalid usage of svcQueryIoMapping. As far as I understand, it's used to get a pointer so passing `nullptr` to it for the returned pointer parameter does not make sense.

* Fix build issues because of gpio_base_vaddr not being passed

* Make the helper functions internally linked (not a fix but a tiny cleanup)


hexkyz:
2 of these issues were already fixed offline for the next commit, but there's no harm in pushing this PR anyway.

PR #93, opened 2018-05-09
Fix stage2_args_t mismatch

encounter:
fixes a couple bugs introduced by e8306361f06132fc1ecc14070bd77c2b4e564dcd

~stage2 is only receiving the first 18 bytes of the bct, don't know what's going on there yet~


TuxSH:
Thanks :)

PR #95, opened 2018-05-10
Implement splash screen loading for fusee stage2

Sorixelle:
Currently, it only supports BMP splashes. If the splash couldn't be loaded for whatever reason, it will fall back to the default splash, statically linked into the stage2 binary.

As the default splash is quite large, this also increases the size of stage2 by quite a bit (~250KB -> >3MB). High speed SD reading might want to be looked into soon, as stage2 takes about 2-3 minutes in my experience to load now.

Example image:
Splash Example


TuxSH:
Will merge later (will resolve conflict if there end up being some).

fusee/fusee-secondary/src/main.c
```diff
@@ -77,7 +78,7 @@ int main(int argc, void **argv) {
 
     g_do_nxboot = loader_ctx->chainload_entrypoint == 0;
     if (g_do_nxboot) {
-        nxboot_main();
+        nxboot_main((uint8_t*)g_framebuffer);
```
jowadmax: g_framebuffer is not defined in this file. This will cause a build error.
Do we need to just add `static void *g_framebuffer = (void *)0xC0000000;` to the file? or should the display also get initialized with it in a similar manner to fusee-primary?

Sorixelle: When I made this PR, `g_framebuffer` was defined there. However, it was moved between then and now, and I did an update to that fork which broke this (probably a bad idea). Tux has mentioned he'll merge and make any necessary changes, however, and the team is aware of this issue.

jowadmax: Ah makes sense. The reason I asked is because I sent a PR to your fork adding PNG support and reducing the Fusee stage 2 payload size to ~470kb.

I was hoping somebody with hardware on hand can test it. I tested as much as I could of it on my host machine, but there's always the possibility that something might not work on real hardware.

Sorixelle: Right now, I think we're undecided on if we want to support multiple file formats, so I just stuck to BMP for now. I think support for other file formats has been brought up before, but we've just said to stick with BMP for the time being. I agree it might be a good idea to help shrink down stage2's size, though. Maybe get in touch with team and see what they feel about other image formats :)

jowadmax: Cool. Yeah my main concern was the boot time. Any millisecond that can be shaved from the boot process will be very valuable.

The PR is available for grabs if the team decided to give it a shot.
I know they're extremely busy now with much more important stuff, but I thought I'd help with what I can :)

CTCaer:
This should be changed into using a 32bit (ARGB) bmp.
And the best way to show it, is to initialize the screen to ARGB instead of RGBA.
This way you can retrieve 32bit values and there's not need to swap Red with Blue.
The loop will have more than 5 times less instructions.

Performance wise, we are talking about a 12000% improvement (3s -> 250ms).

fusee/fusee-primary/src/main.c
```diff
@@ -15,6 +15,7 @@ extern void (*__program_exit_callback)(int rc);
 static void *g_framebuffer;
 static char g_bct0_buffer[BCTO_MAX_SIZE];
 
+#define I_KNOW_WHAT_I_AM_DOING
```
tumGER: I know that this isn't important but you should probably remove that

jowadmax:
@CTCaer wow what an impressive load time boost!

Side question. Have you tried using PNG as the source file? Embedding a PNG as opposed to BMP reduced the Fusee file size from 3mb to ~470kb.

The PR of interest can be found here: https://github.com/Sorixelle/Atmosphere/pull/1

It has very minor changes over the current implementation (+ adding a 1-file PNG library file).

CTCaer:
UPNG messes with something the last time I used it in hekate. So I went the LZ way to compress the bootlogo and never tested it. No time for debugging.

IMO, the file size does not matter a lot because it's loaded from sd card.
What matters the most for a loader is the speed. 

In this case, I would even remove the BMP header or even better save it as RAW with the framebuffer's color channels order. This way a memcpy to remove the header to have aligned access (for performance) is not needed.
And I would let the bmp/PNG rendering (with the proposed changes, depending on file format) for custom ones.

Sorixelle:
In light of recent community events, I've decided to refrain from making any contributions to Atmosphere and other ReSwitched projects. As such, I'm closing this PR.

fennectech:
What’s happened?    Can someone link me?

jowadmax:
https://twitter.com/reswitchedteam/status/1017546918317047808?s=19

https://twitter.com/ktemkin/status/1017606037690433536?s=19

..and others. Unfortunate events fueled by lack of communication and rushed actions.

GGLinnk:
> ktemkin have submitted a bootrom bug she did not originally discover (without finder consent) to bug bounty (for up to $200k).

 Originally Writed by SciresM on the ReSwitched Discord Server.

drofdoom:
What ever happened to the hack in shadow types which bootrom bug ? And with out flashing the chip discovery will do them no good besides trust I dont see an issue just ban and move on by the way they've all sold us out thats why ninty is banning people cause they told them what to look for "cleared logs" layered fs sx os you think they can hack in the open without consequences for nothing in return ask MrHotz how that works .

fennectech:
I refuse to beleve that reswitched sold out the community.   Ninty is watching our actions closely on their own.     I could easily have figured out that cleared logs is a sign of tampering on my own.  And so can any ninty dev.

fennectech:
Hell.  Any normal person could see that cleared logs is a sign of tampering and suggested it to ninty

drofdoom:
Nx report leaves traces in the log I was a temp at infinityward in the early 2000s wii era i used to patch gecko codes .Beleave what you want unfinished hacks getting released in the wild is what did it weither theres or not they weren't ready .

fennectech:
F-G was going to get patched ether way.      And for us already owning switches it doesnt mater
  All these unfinished hacks did is bring it a few months earleyer.

drofdoom:
Nevermind your narrow mind has missed and dulled the point as well you cannot without open communication hack in the public eye they blatantly stated that they were waiting out an nda with nintendo someone leaked it and killed that if they didnt disclose anything woth ninty why the nda .

fennectech:
Wow.   Somebody clearly doesnt understand technology other than how to use a controller.

drofdoom:
Its all on their discord go read .

Sorixelle:
Not to minimod, but this discussion is unrelated to this PR, and the PR has been closed, anyways. You'll probably be best taking this to the ReSwitched Discord server.

PR #97, opened 2018-05-13
Allow for untracked config.h

Qyriad:
`#include "src/config.h"` if it exists in fusée primary

### DIFF
--- a/fusee/fusee-primary/Makefile
+++ b/fusee/fusee-primary/Makefile
@@ -30,6 +30,10 @@ CFLAGS = \
 	-fstrict-volatile-bitfields \
 	-DFUSEE_STAGE1_SRC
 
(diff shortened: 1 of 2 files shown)


kitlith:
couldn't this also be accomplished with gcc's `-include` command line argument? e.g. conditionally add `-include src/config.h` instead of `-DHAS_CONFIG_H`?

jowadmax:
@kitlith yes, but then you force the file to be included with every compiled unit (.c file). The way it's done in this PR makes it so that you can selectively include it in files of your choosing (exclusively) if it the config file happens to exist.

kitlith:
fair enough.

Qyriad:
Due to recent events regarding the ReSwitched team, I no longer desire to support Atmosphère.

PR #108, opened 2018-05-18
figured this may be handy

Azure-Agst:
just updated the readme.

may look familiar, just adapted from the error in fusee's main.c.


hexkyz:
No need to clutter the main README file for this warning which will eventually become irrelevant.

PR #114, opened 2018-05-26
Set SDMMC controller to SDR104 as a workaround

desowin:
According to Tegra X1 Series Processors Silicon Errata there is possible
misalignment of received data which results in a CRC error. The issue is
present only in SDR50 mode.

I didn't check this code. I wanted to start discussion about this on the discord, however I don't have permissions to post to the development channel. Hence I decided a pull request is going to be best way to get it mentioned/discussed.

I didn't yet traverse through the full datasheet nor the driver code, so the change might break something.

### DIFF
--- a/fusee/fusee-primary/src/sdmmc.c
+++ b/fusee/fusee-primary/src/sdmmc.c
@@ -1360,7 +1360,8 @@ static int sdmmc_apply_clock_speed(struct mmc *mmc, enum sdmmc_bus_speed speed, 
         case SDMMC_SPEED_SDR50:
             mmc->regs->host_control |= MMC_HOST_ENABLE_HIGH_SPEED;
             mmc->configure_clock(mmc, MMC_CLOCK_SOURCE_SDR50, MMC_CLOCK_DIVIDER_SDR50, MMC_CLOCK_CONTROL_FREQUENCY_PASSTHROUGH);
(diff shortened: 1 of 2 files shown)


CTCaer:
The code tries to use SDR50 as a fallback, if sth is wrong with switching to SDR104.
So, if SDR104 failed and SDR50 has misalignment errors, shouldn't it instead fallback to SDR25?

desowin:
@CTCaer The errata mentions that the workardound is to set SDR104 in SDMMC controller even if the card is at SDR50.

CTCaer:
@desowin I see. So the plan is to try one more time to enter SDR104 before falling back to SDR25.
Fair enough.

desowin:
@CTCaer This is not exactly the "one more time to enter SDR104". Notice that the clocks are configured according to the SDR50. The only change is the selected UHS mode.

CTCaer:
@desowin 
True, haven't seen this important detail `tuning_attempts = MMC_VENDOR_TUNING_TRIES_SDR50;`
Thanks for letting me know the whole procedure.

[TX1 Errata](https://developer.nvidia.com/embedded/dlc/tx1-errata)

hexkyz:
I've verified that this is indeed the recommended workaround suggested for TX1 hardware.

PR #117, opened 2018-06-02
Use "final" and "override" where appropriate

neobrain:
These keywords were introduced in C++11 for the sake of catching bugs and improving code expressiveness:

* "override" ensures the given member function actually overrides one of the base class's member functions (which can be easy to mess up by picking the wrong types or function names)
* "final" classes may not be inherited from; this implies that none of its child methods will ever be overridden. This is useful when browsing through the source (easy to identify concrete classes that will be instantiated), and it also helps the compiler optimizer at devirtualization.


thomleg50:
there are an error in pm_boot_mode.hpp that done it unable to compile.

PR #130, opened 2018-06-07
PM: Remove magic numbers for flags and events to improve readability and understandability

misson20000:
This replaces the magic numbers used to interact with `Registration::Process.flags` with a set of enum constants. I would appreciate feedback on the names of these constants, as well as feedback on what to do about the notes I made when trying to figure out the meaning of each flag and event type. Should I keep the notes the way they are, remove them, or try to clean them up?

~~I would also appreciate if someone could verify that I didn't break any of the logic accidentally (probably by either accidentally replacing a flag with another one, or by changing the logic for flags 0x10 and 0x20), since I can't test on hardware at the moment.~~ Tested working on 3.0.0.

### DIFF
--- a/stratosphere/pm/source/pm_registration.cpp
+++ b/stratosphere/pm/source/pm_registration.cpp
@@ -105,16 +105,16 @@ void Registration::HandleProcessLaunch() {
         
     /* Setup process flags. */
     if (program_info.application_type & 1) {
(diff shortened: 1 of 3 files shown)


SapphireExile:
From a quick read through, it all seems in order. I'll test on hardware some time soon.

PR #131, opened 2018-06-07
Add component overview to the readme

neobrain:
Loosely based on https://github.com/Atmosphere-NX/Atmosphere/projects - such an overview was really needed, and now the information is in the place where people will actually find it :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,17 @@
 
 Atmosphère is a work-in-progress customized firmware for the Nintendo Switch.
 
+Components
+=====
+
+Atmosphère consists of multiple components, each of which replaces/modifies a different component of the system:
+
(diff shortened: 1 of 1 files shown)


SciresM:
This is a really good thing to add, thanks. The *dumbest* request -- could you invert the order of the component listing? Once that's done will merge.

neobrain:
Sure, will do in a bit!

jowadmax:
It would also be nice if this new section followed the formatting (bold for name of component, indentation, etc) of the Credits section. Not super critical but makes the frontpage look more consistent :)

Thanks for the effort.

neobrain:
Changed the component order by request of @SciresM.

@jowadmax I like the current formatting a bit better, but do feel free to submit a followup PR if you like after this gets merged.

PR #132, opened 2018-06-07
Add sdmmc_dump_regs function

tslater2006:
Allows you to print out all of the tegra sdmmc registers.


tslater2006:
I wasn't aware of that function. I'll change it and update this.

fusee/fusee-primary/src/sdmmc.c
```diff
@@ -3615,3 +3615,37 @@ bool sdmmc_card_present(struct mmc *mmc)
...
     return mmc->card_present(mmc);
 }
+
+/**
+ * Prints out all of the tegra_mmc struct fields
+ *
+ * @mmc mmc The controller with which to dump registers.
+ */
+void sdmmc_dump_regs(struct mmc *mmc) {
+    printk("block_size: 0x%04X\n",mmc->regs->block_size);
```
desowin: To keep the same file contents between primary and secondary, it is possible to:
1. #include <inttypes.h>
2. Change the format specifier to use the macros from inttypes.h, for example:
printk("block_size: 0x%04" PRIX16 "\n",mmc->regs->block_size);

For the 32-bit use PRIX32

tslater2006:
I believe I've updated the code to address both concerns.

hexkyz:
Looks and clean and is a useful function to have around while debugging.

PR #134, opened 2018-06-09
Fix race conditions and misconfiguration in sdmmc

desowin:
Properly configure pull up and pull down offsets for autocal.
Run autocal prior to every transfer.

Prevent race conditions in sdmmc_wait_for_event() - make sure the fault
handler has highest priority, then the target irq, state conditions and
finally the error mask.

Do not clear all bits (|=) when acknowledging fault conditions,
only acknowledge the fault conditions itself.

Enable interrupts before preparing command registers - if sdmmc is fast
enough it can actually finish transfer before we enabled the interrupts.
Enabling interrupts clears the COMMAND COMPLETE status bit.

Temporarily print all the sdmmc messages in stage2 - for yet unknown
reason respecting the log level results in some failures.

This results in working microsd card in stage2 on my switch with Samsung
EVO+ 256GB microsd card.

### DIFF
--- a/fusee/fusee-primary/src/sdmmc.c
+++ b/fusee/fusee-primary/src/sdmmc.c
@@ -910,32 +910,38 @@ static int sdmmc_set_up_clocking_parameters(struct mmc *mmc, enum sdmmc_bus_volt
 {
     // Clear the I/O conditioning constants.
     mmc->regs->vendor_clock_cntrl &= ~(MMC_CLOCK_TRIM_MASK | MMC_CLOCK_TAP_MASK);
-    mmc->regs->auto_cal_config &= ~MMC_AUTOCAL_PDPU_CONFIG_MASK;
 
     // Per the TRM, set the PADPIPE clock enable.
     mmc->regs->vendor_clock_cntrl |= MMC_CLOCK_PADPIPE_CLKEN_OVERRIDE;
-
-    switch (operating_voltage) {
-        case MMC_VOLTAGE_1V8:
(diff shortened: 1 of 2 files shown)


langerhans:
Working here too on three cards:
* Samsungf EVO+ 128GB U1
* SanDisk Extreme 64GB U3
* Noname 1GB

fusee/fusee-primary/src/sdmmc.c
```diff
@@ -910,32 +910,38 @@ static int sdmmc_set_up_clocking_parameters(struct mmc *mmc, enum sdmmc_bus_volt
...
     // Constants above and procedure below from the TRM.
     switch (mmc->controller) {
         case SWITCH_EMMC:
+            if (operating_voltage != MMC_VOLTAGE_1V8) {
+                mmc_print(mmc, "ERROR: eMMC can only run at 1V8, but mmc struct claims voltage %d", operating_voltage);
+                return EINVAL;
+            }
+
+            mmc->regs->auto_cal_config &= ~MMC_AUTOCAL_PDPU_CONFIG_MASK;
+            mmc->regs->auto_cal_config = MMC_AUTOCAL_PDPU_SDMMC4_1V8;
```
langerhans: Shouldn't this be `|=`?

desowin: Abosultely! Will fix soon.

hexkyz:
Confirmed to be accurate. Thank you for this.

PR #144, opened 2018-06-16
Modernize C++ usage

neobrain:
Only touched a couple of instances for each of the following, but it's a start and it might be some inspiration on how new code could be written:

* Prefer algorithms over raw loops
* Prefer RAII (`unique_ptr`/`vector`) over raw pointers
* Prefer `enum class` over a C-style `enum` in cases where the enum values where prefixed with the enum type name anyway
* Prefer `std::variant` over explicitly tagged unions
* Prefer `default` constructors/destructors over empty ones
* Prefer `emplace_back` over `push_back` for initialization `vector` elements inplace rather than by copy
* Prefer direct member initialization over initialization in the constructor
* Prefer `std::array` over C-style arrays

Other than modernizing the codebase, this actually fixes at least some bugs where `delete` was used instead of `delete[]`!

### DIFF
--- a/stratosphere/fs_mitm/source/fs_istorage.hpp
+++ b/stratosphere/fs_mitm/source/fs_istorage.hpp
@@ -5,13 +5,13 @@
 
 #include "debug.hpp"
 
-enum FsIStorageCmd {
-    FsIStorage_Cmd_Read = 0,
-    FsIStorage_Cmd_Write = 1,
-    FsIStorage_Cmd_Flush = 2,
-    FsIStorage_Cmd_SetSize = 3,
-    FsIStorage_Cmd_GetSize = 4,
-    FsIStorage_Cmd_OperateRange = 5,
+enum class FsIStorageCmd {
+    Read = 0,
+    Write = 1,
+    Flush = 2,
+    SetSize = 3,
+    GetSize = 4,
+    OperateRange = 5,
 };
 
 class IStorage {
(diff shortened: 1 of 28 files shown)


neobrain:
FWIW, before merging this somebody with an actual Atmosphere setup should give it a try on hardware - I couldn't really test this other than building every single commit. At least if anything breaks, we'll have an easy time pinning down which specific change it was!

hexkyz:
I'll give it a spin on hardware. Waiting for others' thoughts/opinions before merging.

fincs:
Looks like this PR doubles down on zero-cost abstractions, which is a good thing. However I'm fearing the usage of `std::variant` involves allocating objects on heap and/or RTTI overhead as opposed to normal C++ unions, which don't have any of these hidden costs. Please correct me if I'm wrong.

neobrain:
@fincs: `std::variant` is indeed zero-overhead here, i.e. it doesn't touch the heap. Internally it's just a union + a type index, which can be verified if you e.g. add a `char x[512];` member to `RomFSLooseSourceInfo` and then observe that `sizeof(InfoVariant) > 512` (implying all objects are stack-managed).

fincs:
I see. Thank you for your clarification :)

einsweniger:
Hello there! (I was just browsing, this drew my interest :D)

@neobrain would you mind expanding on why you went with a template for the predicate instead of a lambda (just curious)? I just spent a few minutes until I figured out what it does. I'd love to see some documentation on the `meta_tools.hpp`, to improve maintainability. Otherwise the additions seem great!

neobrain:
@einsweniger Originally, I was trying to have some more flexibility with how `C&` and `C&&` are distinguished as well as having an explicit name for this in the first place, however indeed this isn't necessary anymore now. I'll change it to a lambda instead!

As for the remaining code in meta_tools, is there anything specifically you struggled with that needs to be documented?

einsweniger:
@neobrain one problem I had were the reordered parameters in line 35-36, I just glossed over it without recognizing the changed order. 

The rest seems fine, as long as others know the relation between struct operators and `std::function`, I suspect most do? (This was one of the things that puzzled me most when I started with C++, not sure whats more idiomatic :D)

Brandon-T:
That move construction over copy construction for the shared_ptr is a ton faster than previously because the shared_ptr doesn't have to increase/decrease its atomic count.. Instead it steals it via the move.. :)

SciresM:
Will review tonight, and merge unless I find any glaring issues -- thanks! :)

PR #145, opened 2018-06-17
fusee: Add 4.1.0 kernel patch definitions

Traiver:
Needs tests, can't confirm if it's working properly


hexkyz:
Working after the latest fixes.

PR #146, opened 2018-06-17
Fix compilation

Frogomeli:
On my Windows 10 with the last DevkitPRO, i have errors :

 error: redefinition of 'g_kernel_patch_vers_proc_id_send'
 error: redefinition of 'g_kernel_hook_vers_proc_id_send'

PR #147, opened 2018-06-18
Fix a few issues in the secondary kernel_patches.c file

TheDgtl:
Ran into two issues in the kernel_patches.c file for fusee_secondary, this fixes them.

- Macro wasn't using the "vers" variable, was instead adding the text "vers_"
- "search_pattern" was always failing, comparing it to the one in Luma3DS, the comparison was different. This resolved it.


kwsch:
Note: Macro change is the same as present in #146

hexkyz:
Merging this one in favour of #146 due to the additional search_pattern fix.

PR #152, opened 2018-06-22
point to ctcaer hekate instead of nwert bc. repo deleted

ThatNerdyPikachu:


gerd2002:
To be completely honest, I don't think this change should happen. The original code was authored by naehrwert and not CTCaer, and linking CTCear's repo might be misleading in that the hwinit code was created by him.

romracer:
@gerd2002 I agree with your comment, but perhaps there is some middle ground here.  Something akin to "naehrwert for the hekate project and its hwinit code base.  archived copy [here]."  Not unlike how Wikipedia might link to an archive.org copy of a web article.

Someone could even go so far as to upload a repo and tag a branch on his last commit (if we know it) and link to that directly, such that future commits to some repo are not confused as original work.

CTCaer:
Indeed I didn't work on any of the initial hwinit or hekate stuff.
So I also believe that my name should not be there.

The last commit in upstream was this one: https://github.com/CTCaer/hekate/tree/694a8b539506c67eb217c3febb5e745813c218ad

gerd2002:
@CTCaer wrong place to ask, but could you tag the last upstream release in your repo?

hexkyz:
@gerd2002 and @CTCaer are right. The code was based on the original version of hekate which is no longer public.

Additionally, for improved future maintainability and solidarity with the author, I'm writing hekate's hardware initialization code from scratch for a proper integration with Atmosphère.
It's intended to still be inspired by the original hekate while being complemented by re-interpretation and reverse engineering of the hardware bring-up code, so it makes sense to keep the original credit message despite the repo no longer being available.

PR #153, opened 2018-06-23
fusee: Allow declaring yourself knowledgeable via an environment variable

neobrain:
I couldn't find any method of declaring myself knowledgeable without dirtying the git working directory, which makes rebases a pain (among other things git refuses to do with a dirty tree).

This change allows compiling things without having to modify the source files or Makefiles.

Tested build with:
* NEW_ENV_VAR=1 make, build passes as expected
* NEW_ENV_VAR=0 make, build fails as expected
* (NEW_ENV_VAR unset) make, build fails as expected

### DIFF
--- a/fusee/fusee-primary/Makefile
+++ b/fusee/fusee-primary/Makefile
@@ -28,6 +28,10 @@ INCLUDES	:=	include
 ARCH	:=	-march=armv4t -mtune=arm7tdmi -mthumb -mthumb-interwork
 DEFINES :=	-D__BPMP__ -DFUSEE_STAGE1_SRC
 
+ifeq ($(FUSEE_I_KNOW_WHAT_I_AM_DOING),1)
+DEFINES +=	-DI_KNOW_WHAT_I_AM_DOING
+endif
+
 CFLAGS	:= \
 	-g \
 	-O2 \


fincs:
Instead of adding workarounds to make `I_KNOW_WHAT_I_AM_DOING` settable through an env var, how about removing it completely altogether? This mechanism makes an unmodified Atmosphère source tree unbuildable without source code modifications, and its intent purpose of keeping random people from building Atmosphère has completely failed, and there are already countless bastardised and repackaged versions of its components.

Benoit7413:
The error is clearly explicit after all and some commit were pushed on forks with those defines

hexkyz:
I vouch for @fincs suggestion. Seems pretty pointless to maintain the check at this point.

jordanhenderson:
I would recommend removing this mechanism altogether, it seems like a moot point. I would say it's a reasonable assumption that anyone game enough to attempt compilation is implicitly aware that the SW is not release quality just yet, this just adds an extra step for anyone interested in the project.

Benoit7413:
@jordanhenderson You even need to get the last libnx because the one included in packages is out of date

hexkyz:
In the process of preparing for an official release, this check has now been removed.

PR #156, opened 2018-06-26
Replace NAND by eMMC

Benoit7413:
Using NAND doens't really make sense


hexkyz:
NAND and eMMC are interchangeable terms in this context, but since we do use eMMC pretty much everywhere I guess this is fine.

PR #158, opened 2018-06-27
Stage 1 sdmmc.c update log levels

Benoit7413:
error, info and debug levels (0, 1, 2)

As long as we are using 2 and 3 log level nothing should change (just in case we implement UART logs and ini file log levels / importants must always be displayed)

### DIFF
--- a/fusee/fusee-primary/src/sdmmc.c
+++ b/fusee/fusee-primary/src/sdmmc.c
@@ -686,20 +686,33 @@ static void mmc_print(struct mmc *mmc, char *fmt, ...)
     va_list list;
 
     va_start(list, fmt);
-    mmc_vprint(mmc, fmt, 0, list);
+    mmc_vprint(mmc, fmt, 1, list);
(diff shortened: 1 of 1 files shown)


Benoit7413:
Sure code consistency is important, but if files are exactly the same (I didn't checked) we can create a common directory and use the same files for both stage 1 and stage 2 

I was thinking using a function like 
`void console_print(int req_level, ...) `
then extract the str with va_args 
Copying a functions with different levels is useless currently it's better to invoke somethong like this `console_print(WARN, "Blabla %s", somevariable)` and get a result [WARN] Blabla value content

TuxSH:
> Sure code consistency is important, but if files are exactly the same (I didn't checked) we can create a common directory and use the same files for both stage 1 and stage 2

It'll be refactored later.

hexkyz:
The SDMMC driver has been re-designed in order to fix multiple issues. These logging levels have also been included.

PR #169, opened 2018-07-29
Define wait() in fusee-primary/secondary

encounter:
fusee-primary and secondary were linking to sys/wait.h, which didn't quite work :slightly_smiling_face: 

Unrelated: I'm noticing (on my 4.0.1 system) exosphere's bootup_misc_mmio is freezing the fusee-secondary. I can add a panic before `sync_with_nx_bootloader(NX_BOOTLOADER_STATE_FINISHED_4X);` to verify that exosphere continues execution, but fusee-secondary freezes before it can get `SECMON_AWAKE`. Not sure how I can help debug further, or if this is already known.

### DIFF
--- a/fusee/fusee-primary/src/utils.c
+++ b/fusee/fusee-primary/src/utils.c
@@ -13,6 +13,13 @@
 
 #include <inttypes.h>
 
+void wait(uint32_t microseconds) {
+    uint32_t old_time = TIMERUS_CNTR_1US_0;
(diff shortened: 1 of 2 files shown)


SciresM:
Much appreciated :)

By the way, that's odd -- the BPMP should be alive until Exosphere halts it explicitly in setup_4x_mmio: https://github.com/Atmosphere-NX/Atmosphere/blob/master/exosphere/src/bootup.c#L244

PR #171, opened 2018-07-30
fusee-secondary: Correct else-if condition in xmemmove()

lioncash:
Previously both the if and else conditions were the same.

Probably needs an extra set of eyes on this, just to ensure the right branch is being changed here.


SciresM:
lgtm

PR #193, opened 2018-08-19
Change CondVar API to latest version

RKX1209:
See: https://github.com/switchbrew/libnx/commit/1e349b6ce8cb1813091b7b58a43c7a9697775b4f

### DIFF
--- a/stratosphere/libstratosphere/include/stratosphere/hossynch.hpp
+++ b/stratosphere/libstratosphere/include/stratosphere/hossynch.hpp
@@ -8,15 +8,15 @@ class HosMutex {
         HosMutex() {
             mutexInit(&this->m);
         }
-        
+
         void lock() {
             mutexLock(&this->m);
         }
-        
+
         void unlock() {
             mutexUnlock(&this->m);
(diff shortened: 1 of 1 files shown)


fincs:
This PR should be merged when the next stable libnx version gets released, but not before.

RKX1209:
Sure. If this API changes still remain after next libnx release, plz merge it.

fincs:
The API changes are going to remain. It was changed to match other popular condvar APIs (including standard C threads, standard C++ threads, pthreads and Win32 threads), it also makes it easier to implement them in the future. Also, I think this condvar wrapper is conceptually broken. The mutex used alongside condvars isn't supposed to be abstracted away, you're supposed to surround critical code that might use the condvar with appropriate Lock/Unlock. In fact, you *need* to lock the mutex before waiting or signalling a condvar.

RKX1209:
OK

SciresM:
Ended up fixing CondVar class per @fincs instruction.

Thank you very much, though, @RKX1209 :)

PR #196, opened 2018-08-21
[WIP] Awful (and not tested) Multi-User LFS support

ThatNerdyPikachu:
If anyone could test this, that'd be great.

Code is awful, but it should work.

TODO:
- [ ] Add error handling

### DIFF
--- a/stratosphere/fs_mitm/source/fsmitm_utils.cpp
+++ b/stratosphere/fs_mitm/source/fsmitm_utils.cpp
@@ -2,6 +2,7 @@
 #include <stratosphere.hpp>
 #include <atomic>
 #include <algorithm>
+#include <dirent.h>
 
 #include "sm_mitm.h"
 #include "debug.hpp"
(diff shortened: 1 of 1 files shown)


SciresM:
First off, thanks for the PR! :)

Definitely test it -- the code you linked won't actually work, for example -- "opendir" uses libnx's devoptabs, which fs.mitm doesn't use (and can't, due to intentionally never calling fsdevMountSdmc()).

I also have concerns about whether the account initialization stuff -- are you entirely certain account init won't cause a hang due to dependence on FS? There are some issues with other services and this problem.

Also, it looks like your paths are inconsistent -- you use check for the presence of `titles/%016lx_%s` but actually mount `titles/%016lx_$%s`, and you only modified one of several SD access functions that are used for RomFS content querying. Fwiw, I wouldn't use either of those two -- the titles/ dir should never have anything other than title id subfolders. 

Anyway, for multi-user layeredfs, it seems to me like the ideal layout looks like /titles/%016x/romfs_<user id> if and only if present, falling back on /titles/%016x/romfs if no matching user is found. I'm not sure if user profile names are unique -- if they're not, I'm not sure if we can actually use them.

ThatNerdyPikachu:
Alright, hm.
- If acc depends on FS, it won't work, correct?
- What functions are ideal for me to modify to create the indented effect?
- The feature request specified the syntax of the folders (tid_user), and once the initial code got approved, I was going to

yellows8:
Probably shouldn't (?) assume the account service is already registered either.

Those '$' should be '%' in snprintf, and use sizeof()-1.

ThatNerdyPikachu:
@SciresM Er, how many file openers mention the tid folder?

AnalogMan151:
https://github.com/Atmosphere-NX/Atmosphere/search?q=%2Fatmosphere%2Ftitles%2F%25016lx&unscoped_q=%2Fatmosphere%2Ftitles%2F%25016lx

PR #199, opened 2018-08-25
fusee-primary: Fix always false conditional in xmemmove

lioncash:
This was corrected in 4b8455baf91d8699a663341c135c91c34f0a7816 for fusee-secondary, but I neglected to also fix it in fusee-primary (my bad!)

### DIFF
--- a/fusee/fusee-primary/src/chainloader.c
+++ b/fusee/fusee-primary/src/chainloader.c
@@ -17,7 +17,7 @@ static void *xmemmove(void *dst, const void *src, size_t len)
         for (size_t i = 0; i < len; i++) {
             dst8[i] = src8[i];
         }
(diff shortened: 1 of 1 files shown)


CTCaer:
That's because nxbootloader effectively does this.
`MEMORY[0x7000E4E8] = 2 * (MEMORY[0x7000E4E8] >> 1) | MEMORY[0x7000E4E8] & 1`

Which can be translated to:
`MEMORY[0x7000E4E8] = (MEMORY[0x7000E4E8] & 0xFFFFFFFE) | MEMORY[0x7000E4E8] & 1`
or just simply:
`MEMORY[0x7000E4E8] = MEMORY[0x7000E4E8]`

Issue #206, opened 2018-09-18
Improvement: improve fusée screen printing


Resaec:
For 1.0 or later a rework of the current printf/k to a single function to include all message types except fatal_error would be nice.

Supplying a type of message (DEBUG, INFO, MANDATORY, WARNING) to the function would apply colloring as well as prefixes like `[DEBUG] ` to the messages print to screen.

Maybe it could also hide the splash screen in case of type > `MANDATORY` so the user can see that something went wrong as long as there is no SD logging in fusée?

This would also enable users to define the log level in BCT.ini and reduce DEBUG clutter on screen.
Printing to screen is probably quite slow, so this would further improve the boot time.

As printf/k use va_list it would be easiest to make `messageType` the first parameter of this new function.

Please share your thoughts.

hexkyz:
I agree, this is exactly something I had envisioned. Now that fusee is functional, I believe it would be preferable for it to be silent (displaying only the logo) unless specifically told to show log messages.

hexkyz:
The referenced PR has been merged. I'll work on adding support for choosing logging levels with the BCT.ini file and then this issue can probably be closed.

hexkyz:
Implemented logging level configuration from BCT.ini in the latest commit. Closing this now.

Issue #182, opened 2018-08-02
Clean up Fusee UI for 0.7 release.


SciresM:

ThatNerdyPikachu:
@SciresM What exactly do you need to clean up in it? I'd be happy to help with cleaning it up!

HylianMedia:
@SciresM @ThatNerdyPikachu You guys need any extra help with this? I'd be glad to lend a hand if needed.

hexkyz:
Splash screen is now available as part of fusee. Please discuss here what other features/cleanup would be desirable for release.

hexkyz:
Fusee UI's clean up has now been concluded.

PR #209, opened 2018-09-23
Silenced boot from stage 2

Resaec:
I wrote a few functions implementing silenced booting from stage 2 as suggested and aggreed on in #206 
It boots, and lets you set the required log level for a message to be displayed on screen.
There are placeholders for UART logging all messages with prefixes.
As i could not find out how the BCT.ini parsing had to be changed for me to implement a parameter there I left that for later.


Resaec:
This could also be seen as part of #182

hexkyz:
Awesome! We just need to sort out some design choices before merging:
- Consider getting rid of "driver_utils.h" completely;
- Port centralized logging to fusee-primary (which uses separate printk and vsprintf source files);
- Discuss if log.c/log.h should be under lib/ folder or not.

cc @SciresM @TuxSH

Resaec:
I think log should go to lib/ too. If the others disagree I will change it back later.
Printing is ported and useable on stage 1 and 2 and `lib/driver_utils.h`has been removed.

Using `vsprintf.h` in fusee-primary code is mandatory (and the better choice?) as the payload gets bloated by `stdio.h` to the point where it wouldn't fit anymore :)
Same with `video_fb.h` which provides `video_puts()`.

In fusee-secondary I get a strage bug, where the screen wouldn't turn on for stage 2 and the splash if I don't include `stdio.h`.

PR #217, opened 2018-10-05
pm: add AtmosphereGetCurrentLimitInfo command to pm:dmnt

misson20000:
Tested on 5.0.0.

### DIFF
--- a/stratosphere/pm/source/pm_debug_monitor.cpp
+++ b/stratosphere/pm/source/pm_debug_monitor.cpp
@@ -17,6 +17,7 @@
 #include <switch.h>
 #include <stratosphere.hpp>
 #include "pm_registration.hpp"
+#include "pm_resource_limits.hpp"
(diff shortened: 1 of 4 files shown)


stratosphere/pm/source/pm_debug_monitor.cpp
```diff
@@ -158,3 +165,27 @@ std::tuple<Result, CopiedHandle> DebugMonitorService::get_process_handle(u64 pid
     }
     return {0, proc->handle};
 }
+
+std::tuple<Result, u64, u64> DebugMonitorService::get_current_limit_info(u32 category, u32 resource) {
+    if(category > ResourceLimitUtils::ResourceLimitCategory::ResourceLimitCategory_Applet) {
+        return {0xf001, 0, 0};
```
misson20000: There may be a better result code to use here.

PR #218, opened 2018-10-05
Fix ProcesState enum field names

ThatNerdyPikachu:


fincs:
Should be merged when the next stable libnx is released.

ThatNerdyPikachu:
@fincs If you don't mind me asking, when will that be?

PR #219, opened 2018-10-06
loader: add SetExtraMemory extension

misson20000:
Adds a command to `ldr:shel` to make it so the next time a specific title is loaded, it will have extra code pages that loader leaves alone. Twili (misson20000/twili#52) uses these extra code pages to load an NRO. If anyone else ever wants to implement a multiprocess homebrew loader, this will probably work for that, too.

Tested on 5.0.0.


SciresM:
I'm gonna say no to this one -- I don't actually understand a practical use case for this?

PR #221, opened 2018-10-08
Make sure a title has fs mitm mods/files on the SD before enabling

### LINKED ISSUE #202: fs mitm causes some games to crash upon launch
for example, the messanger is one that i know of.

ThatNerdyPikachu:
This is a temporary workaround for #202, and the code is taken from Rei's fork of Stratosphere, seen here: https://github.com/Reisyukaku/NX_Sysmodules/blob/3c12add04883b893698cfa04451c834e34bcbe66/fs_mitm/source/fsmitm_service.hpp#L24-L33

### DIFF
--- a/stratosphere/fs_mitm/source/fsmitm_service.hpp
+++ b/stratosphere/fs_mitm/source/fsmitm_service.hpp
@@ -38,7 +38,14 @@ class FsMitMService : public IMitMServiceObject {
         }
         
         static bool should_mitm(u64 pid, u64 tid) {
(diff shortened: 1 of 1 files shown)


bakugo:
>This fixes #202 

It's a workaround, not a fix.

ThatNerdyPikachu:
true

tumGER:
Shouldn't you rather cherry-pick the commit on the original repo? This would give credits to the original author instead of a "lol" commit

PR #225, opened 2018-10-11
loader: add SetExternalContentSource extension

misson20000:
Currently WIP, as I've not yet finished writing anything to test this. Posting here for review.

One thing that I'm noticing is that the content management code is becoming very convoluted. Perhaps there should be a different design where there is a stack of content providers that are queried for a given TID+NSO index, where each content provider can either supply the file, reject the file, or pass the request along to the next content provider.

- ExternalContentSource provider overrides everything
- HBL content provider
- SD content provider
- ExeFS content provider

This type of design would also allow a more powerful extension so that rather than using `SetExternalContentSource` to override a specific title's contents, a sysmodule could use a more powerful `AddExternalContentProvider` and use its own logic (rather than the limited logic available in loader) to determine override behavior. I'd like to hear your thoughts on this @SciresM.

### DIFF
--- a/stratosphere/loader/source/ldr_content_management.cpp
+++ b/stratosphere/loader/source/ldr_content_management.cpp
@@ -19,11 +19,12 @@
 #include <strings.h>
 #include <vector>
 #include <algorithm>
+#include <map>
 
 #include "ldr_registration.hpp"
 #include "ldr_content_management.hpp"
 #include "ldr_hid.hpp"
-
+#include "ldr_npdm.hpp"
 
 #include "ini.h"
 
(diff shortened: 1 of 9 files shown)


misson20000:
I've had to change the original design of the command from `SetExternalContentSource(u64 tid, handle<clientsession<IFileSystem>, move>)` to `SetExternalContentSource(u64 tid) -> handle<serversession<IFileSystem>, move>`. This is similar to the design of `AtmosphereInstallMitm` in SM.

misson20000:
Tested working on 5.0.0, for real this time. Some things that are a little bit dirty about this are that since I mount ECS on `code:`, I have to short circuit all the override checks to false for the currently ECS'd title and I have to keep it mounted until process launch completes, since normal process launch flow actually mounts ExeFS twice, but we can't get the ECS-Filesystem handle back once we close it like we can for actual ExeFS.

misson20000:
Changed to mount ECS as `ecs-<title.id>` and use the same content-overriding patterns as HBL/SD. Also now supports up to 16 external content sources at a time.

PR #227, opened 2018-10-17
Adding pattern matching functions

Resaec:
This PR includes functions for pattern matching in static and a dynamic(ASLR) memory searches.

Exported functions are:
memSearch for static target pattern matching, uses memmem.
memSearchASLR for dynamic target pattern matching, using either a binary or a byte array mask.
Both types support skipping to ignore false positives.
The ASLR types also offer a byte alignment parameter which decrease search time.

Part of  #3

I can currently not compile Atmosphere, so I only tested it standalone.

### DIFF
--- a/fusee/fusee-secondary/src/lib/memsearch.cpp
+++ b/fusee/fusee-secondary/src/lib/memsearch.cpp
@@ -1,0 +1,153 @@
+#include "memsearch.h"
+#include <string.h>
+
+bool matchPattern(
+    const uint64_t startLocation,
+    const unsigned char* pattern,
+    const uint32_t patternSize,
+    const uint64_t mask)
+{
+    for (uint32_t i = 0; i < patternSize; i++)
+    {
+        /* if the bit in our mask is 0 this is a wildcard */
+        if ((mask & 1 << i) == 0) continue;
+
+        /* if the byte does not match our pattern we can skip here */
+        if (*((unsigned char*)startLocation + i) != pattern[i]) return false;
+    }
+
+    return true;
+}
+
+uint64_t findPattern(
+    const uint64_t startLocation,
+    const uint64_t searchAreaSize,
+    const unsigned char* pattern,
+    const uint32_t patternSize,
+    const uint64_t mask,
+    unsigned int skipCount,
+    const int alignment)
+{
+    for (uint64_t currentLocation = startLocation; currentLocation + patternSize <= startLocation + searchAreaSize; currentLocation += alignment)
+    {
+        /*
+        If we have aligning code we can skip 2, 4 or 8 bytes each step
+        this will speed up this process significantly!
+        This has to be keept in mind for pattern generation!
+        Patterns can then only be starting at those alignments.
+        */
+
+        if (matchPattern(currentLocation, pattern, patternSize, mask))
+        {
+            /* we found a matching location */
+            /* if we have to skip, do so */
+            if (skipCount)
+            {
+                skipCount--;
+                continue;
+            }
+
+            /* we are done, target found, no more skipping */
+            return currentLocation - startLocation;
+        }
+    }
+
+    return 0;
+}
+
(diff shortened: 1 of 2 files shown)


Resaec:
Sure, pattern scanning is slow.
The speed increases linear with higher byte alingment values. So if you align your pattern to 4 bytes in the binary you want to search it already is pretty fast.

You only use pattern scanning once per pattern and binary lifetime so the speed is negligible in comparison to the potential to survive an update and reshuffeling at loadtime.

"memSearch-like" makes it sound like they are standard. I took it you meant pattern scanning/matching function in general, right?.

m4xw:
> Sure, pattern scanning is slow.
> The speed increases linear with higher byte alingment values. So if you align your pattern to 4 bytes in the binary you want to search it already is pretty fast.
> 
> You only use pattern scanning once per pattern and binary lifetime so the speed is negligible in comparison to the potential to survive an update and reshuffling at loadtime.

Want to make the startup even slower? It bad enough as-is.
Once per lifetime is one too much, if at all it needs to save the patches after the initial run.

> "memSearch-like" makes it sound like they are standard. I took it you meant pattern scanning/matching function in general, right?.

memsearch is as old as programming itself, there are dozens of implementations.
Don't use it. 
Its error prone, absolutely useless without actual heuristics which would be horrible at run-time.
This is **not** the 3DS.

> Both types support skipping to ignore false positives.

You like corruption, don't you?
False positives are not acceptable in such a use-case, like, at all. memsearch is always **just** a _indication_

> If with "DRAM pre-training" you mean caching, yes this code is as cache friendly as it gets with this kind of task.

No, I don't talk about caching at all.
This is about memory training, without that the frequency is extremely low and any access to DRAM is snail speed.
Even worse if you execute instructions from DRAM. i-/d-cache don't help much for this.

Resaec:
All valid points.
But this is not memsearch as it was implemented for 3DS.
Luma3DS's memsearch was only able to scan static binarys, while this is able to scan ASLR memory by implementing a pattern mask.
Sure above application level patterns have to be unique. But for live game patching this is perfectly fine.

I agree, with me placing this in fusee I made it look like it should be used at boot. That was a mistake. But this is meant as a discussion for the time being.

Pattern matching was requested and absolutely has its place. Less in the system than the application level, true, but non the less usefull.
Try to be constructive, instead of bashing about a skip feature. If you don't want to use it or think it is bad in your use case, note that it should not be used there and why.

Thanks for the update on DRAM pre-training.

m4xw:
> Luma3DS's memsearch was only able to scan static binarys, while this is able to scan ASLR memory by implementing a pattern mask.

Doesn't make it any better. Sorry if that sounds non-constructive, but I have a Issue with people of a certain fork-fw advertising it as the holy grail of patching, when in reality its just litter.

> Sure above application level patterns have to be unique. But for live game patching this is perfectly fine.
Honestly, I disagree. Just because it works, doesn't mean its good.

As for that, just disable ASLR altogether, you will still need to update the patch everytime any game gets a update, I bet on that.

> Pattern matching was requested and absolutely has its place. Less in the system than the application level, true, but non the less usefull.

Again, I don't think it has a place - at all.
If other people have different opinions, so be it.

> Try to be constructive, instead of bashing about a skip feature. If you don't want to use it or think it is bad in your use case, note that it should not be used there and why.

This is not bashing, it's just polluting the code-base IMO.

If you want that feature, just add it to your kip or whatever.

My 2 cents

NWPlayer123:
> This is not bashing, it's just polluting the code-base IMO.
> 
> If you want that feature, just add it to your kip or whatever.

This is literally how you make more "fork-fws" so I don't know what you're trying to accomplish here

Resaec:
Is this still a desired feature?
If so, should I rather move it to Exosphère or Stratosphère?
To me Exosphère sounds like the better place as it could be used to find code in OFW modules after updates.
If not, it should be closed after a month of silence.

SciresM:
Closing due to inactivity, and also because I don't actually think there is anything in our current patchset that needs or benefits from pattern matching (even if theoretically it may be beneficial?)

PR #263, opened 2018-11-12
Add build script for use with AppVeyor

gabehxd:
https://www.appveyor.com/

Most of the credit goes to @shadowninja108 for the original version of the script

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,52 @@
+version: AppVeyor build {build}
+image: Ubuntu
+init:
+- sh: >-
+    mkdir ../resources
+
+    mkdir ../libnx
+
+    wget https://github.com/devkitPro/pacman/releases/download/devkitpro-pacman-1.0.1/devkitpro-pacman.deb -P ../resources
+install:
+- sh: >-
+    sudo dpkg -i ../resources/devkitpro-pacman.deb
+
+    echo -en "\n\n" | sudo dkp-pacman -S switch-dev
+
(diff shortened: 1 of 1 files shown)


SciresM:
I'd rather we use docker than appveyor -- I really don't like the potential for desync between Makefile and appveyor script as new modules get added to the Makefile.

I'd suggest we use the [devkitPro docker repo](https://hub.docker.com/r/devkitpro/devkita64_devkitarm).

joel16:
Only "issue" (or perhaps downside is a better word) with devkitPro docker's repo is that it only seems to get updated if devkitA64 gets updated. New libnx versions aren't accounted for, so if you pushed code to libnx and a new version of libnx is released, I don't think the docker file gets updated until devkitA64 is updated as well so you may see a lot of failed builds per commit.

tiliarou:
The docker is indeed not really useful for daily builds since it's far behind from rest of the development...
The few times I tried it to build HB and Hekate it failed and needed manual fix.

joel16:
@fincs it is the truth?

fincs:
> `--env DEVKITARM=/opt/devkitpro/devkitARM`

I don't see why this is necessary. The docker images should already have that env var set internally through the `devkit-env` package.

zaksabeast:
`devkita64_devkitarm` doesn't appear to.

Enter the image with `docker run -it devkitpro/devkita64_devkitarm /bin/bash`, then run `echo $DEVKITARM`.

Alternatively, try building something using the devkita64_devkitarm image that needs to use the DEVKITARM environment variable.

PR #287, opened 2018-12-02
Add a way for users to identify where builds come from.

StevenMattera:
Very small tweak to allow for you to set an environment variable (AMSSRC) when building to add an identifier of the source of the build. It will just add it in brackets next to version numbers in the creport files, fatal screen, and system version identifier. The idea is to help with supporting users to know where their build came from. (I haven't actually tried building this, but the code changes are so simple I'm confident in it working.)

### DIFF
--- a/exosphere/Makefile
+++ b/exosphere/Makefile
@@ -34,6 +34,11 @@ INCLUDES	:=	include ../common/include
 #---------------------------------------------------------------------------------
 ARCH	:=	-march=armv8-a -mtune=cortex-a57 -mgeneral-regs-only #<- important
 DEFINES	:=	-D__CCPLEX__ -DATMOSPHERE_GIT_BRANCH=\"$(AMSBRANCH)\" -DATMOSPHERE_GIT_REV=\"$(AMSREV)\"
+
+ifneq ($(strip $(AMSSRC)),)
+	DEFINES	+= -DATMOSPHERE_BUILD_SOURCE=\"$(strip $(AMSSRC))\"
+endif
+
 CFLAGS	:= \
 	-g \
 	-O2 \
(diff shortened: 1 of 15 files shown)


SciresM:
@StevenMattera There are pretty hard restrictions on versioning strings -- in particular, the string in settings can only have a four character build source under your scheme before the version string is too long. This will be even worse if Atmosphere hits a double digit minor version, or if Nintendo ever hits 10.0.0 (with both, you would only have two characters for a build source).

I don't really see this as viable atm; I'm also very wary of putting arbitrary-length strings in fatal, which considers e.g. string size very carefully.

Can I ask what your imagined use-case is? What do you think AMSSRC would look like? I worry in particular about other builds just copying whatever value I put in, and I'm not wholly convinced that this actually saves any effort in debugging.

StevenMattera:
@SciresM Okay that is good to know. I wasn't aware of the length limit on the versioning string. Honestly that is probably the least important one.

The use-case for crash report and fatal is to know which build someone is using when they are having an issue. The first idea was to change the image, but I felt like that was rebranding and didn't like it. So I suggested just adding a simple string to be able to identify where a build came from. So if a user takes a picture of their fatal screen or send the crash report file we can easily see if they are using a Kosmos build vs a build from Ave's build server vs official builds. So below could be a few examples:

```bash
export AMSSRC=KSM
export AMSSRC=AVE
export AMSSRC=REI # ¯\_(ツ)_/¯
unset AMSSRC # Official builds can continue to leave this blank.
```

Let me experiment with limiting the size of the string, if not in the makefile then in code, and also remove it from `set_mitm`.

SciresM:
I'm gonna be honest -- I don't think we really care about other distributors.

If someone comes to us for support, we can simply ask them to try using official binaries.

PR #302, opened 2018-12-21
sm: add query registrations extension

misson20000:
Adds an extension that can be used to query service registrations. The offset-count pattern is used by Nintendo code (ns:am ListApplicationRecord for example). I've made a new struct called `ServiceRecord` instead of using `Registration::Service` so that we can have a stable API if we need to change `Registration::Service`, but I've also added an output for the size of each record so that old client code can keep working if we want to add more fields to `ServiceRecord`.

Here's the code I tested this with: https://gist.github.com/misson20000/2e22a7a35ce67b87c6bc723a52df2d36. Note that my code is dangerous and doesn't care if `record_size` is smaller than it expects. The example pseudocode in the documentation is better about this.

### DIFF
--- a/docs/modules/sm.md
+++ b/docs/modules/sm.md
@@ -20,6 +20,16 @@ interface nn::sm::detail::IUserInterface is sm: {
 }
 ```
 
+Additionally, several commands have been added to the [`sm:m`](https://reswitched.github.io/SwIPC/ifaces.html#nn::sm::detail::IManagerInterface) interface.
+
+Their SwIPC definitions follow.
+```
+interface nn::sm::detail::IManagerInterface is sm:m {
+  ...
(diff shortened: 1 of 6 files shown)


SciresM:
I think it's seriously worth considering splitting into a sm:dmnt, rather than adding to sm:m. Should evaluate for some stuff I added to sm:m too.

SciresM:
I think it's seriously worth considering splitting into a sm:dmnt, rather than adding to sm:m. Should evaluate for some stuff I added to sm:m too.

misson20000:
Yeah, my initial instinct was to make a new `sm:ams` service for this actually but then I saw that you had already added extensions to `sm:m`.

PR #304, opened 2018-12-21
sm: add permissive unregistration extension

misson20000:
Adds a compile-time check to remove the requirement that a service may only be unregistered by a built-in or the process that registered it.

My use-case for this is so that I can develop a sysmodule replacement by booting the system with the original sysmodule running, then unregistering its services (letting it keep all its sessions), launching my reimplementation (letting it register the services in place of the original sysmodule), and then launching a game.


SciresM:
What is the benefit of this over installing a MitM that does not do any forwarding?

misson20000:
Hmm, I don't think there actually is one. Good point!

PR #306, opened 2018-12-21
stratosphere/Makefile: change KIPS to MODULES

misson20000:
It's a bit confusing that this variable is called `KIPS`, when really only some of them are KIPs.


SciresM:
Yeah, good call.

PR #317, opened 2019-01-06
loader: add 'override any app' feature

natinusala:
This allows overriding any app with HBL by using `hbl_tid=app` in `loader.ini`. Useful with `override_key=R`, to run HBL on top of any game just by pressing R when launching it.

### DIFF
--- a/stratosphere/loader/source/ldr_content_management.cpp
+++ b/stratosphere/loader/source/ldr_content_management.cpp
@@ -40,6 +40,7 @@ static char g_hbl_sd_path[FS_MAX_PATH+1] = "@Sdcard:/atmosphere/hbl.nsp\x00";
 static u64 g_override_key_combination = KEY_R;
 static bool g_override_by_default = true;
 static u64 g_override_hbl_tid = 0x010000000000100D;
+static bool g_override_any_app = false;
 
 /* Static buffer for loader.ini contents at runtime. */
 static char g_config_ini_data[0x800];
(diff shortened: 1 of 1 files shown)


CTCaer:
"unused" buttons or combos, seem more normal to me.
E.g. L3 or R3 (stick buttons)

SciresM:
@CTCaer eh, it seems a bit odd to make a user have to hold down a stick button. Plus stuff should be user-configurable.

natinusala:
Is it good enough to compare the tid with `0100000000010000` ?

CTCaer:
I don't follow why it's odd. The stick is zero finicky and it's used since ancient times with PS2 easter eggs.

True that it should be configurable (although almost no one changes this).
The only problem is that >= 0100000000010000 will not always work for the better.

natinusala:
I changed `*` with `app` and added a title id filter, it needs to be `>= 0100000000010000` to be overriden.

misson20000:
TID filtering seems to be a bit of a hack. Any reason you can't just check the application bit?

natinusala:
I didn't find any documentation on the title id format so I didn't know there was such a bit. 

Which bit is it ? I can submit another PR to improve the filter.

misson20000:
Not in the TID, but in the process's flags from NPDM

SciresM:
@misson20000 I accepted it on two grounds:

1. fs.mitm must also be able to filter on any condition we use, and fs.mitm does not have access to NPDM flags or metadata.

2. The official application TID range is `0x0100000000010000`-`0x01FFFFFFFFFFFFFF`, per game NPDMs. Anything outside of that range is not an official application, and so we don't necessarily need to care about it.

misson20000:
Alright, that makes sense.

PR #321, opened 2019-01-09
loader: add 'override any app' feature

tumGER:


tumGER:
Forget that - Typed the wrong thing there

PR #333, opened 2019-01-25
Add component subdirectories to root Makefile's .PHONY

nicoelayda:
I noticed that calling `make` (or `make clean`) with the root Makefile does not properly build (or clean) troposphere. This should fix that and prevent similar issues in the future.


SciresM:
Thank you! :)

PR #351, opened 2019-02-20
Copy new sept-primary.bin to out folder

tslater2006:


SciresM:
no, not how this will work.

PR #352, opened 2019-02-20
Create sept docs

TheCoder3156:
If there is any technical stuff I missed I'll add it


SciresM:
why do people keep trying to PR stuff for sept before I'm even done with it?

TheCoder3156:
Sorry :(

PR #370, opened 2019-02-22
Don't skip sept-secondary.bin when SEPT_ENC_PATH is set

igoticecream:
Hello,

This PR fixes the `make dist` after a fresh clone or clean when using the prebuild sept-secondary.enc ($SEPT_ENC_PATH)

Currently failing because sept-secondary.bin is not found

### DIFF
--- a/sept/sept-secondary/Makefile
+++ b/sept/sept-secondary/Makefile
@@ -120,8 +120,12 @@ check_rebootstub:
 	@$(MAKE) -C $(AMS)/exosphere/rebootstub all
 
 $(BUILD):
+ifeq ($(strip $(SEPT_ENC_PATH)),)
 	@[ -d $@ ] || mkdir -p $@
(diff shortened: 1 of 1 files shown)


leo60228:
Wouldn't it make more sense to do something like:
```
ifeq ($(strip $(SEPT_ENC_PATH)),)
	@python $(TOPDIR)/sept_sign.py $< $@
else
	@cp $(SEPT_ENC_PATH) $@
endif
@echo built ... $(notdir $@)
```

thedax:
Alternatively, `touch $(TOPDIR)/sept-secondary.bin` instead of moving it and initiating a dummy build. I don't really see a lot of point in actually building a dummy sept-secondary if the the prebuilt one is provided.

igoticecream:
@thedax yea that is a better solution, thanks. I'm going to make another PR

thedax:
Fair enough, and no problem. Apologies if I seemed hostile in my initial message, I didn't mean to come across as such.

igoticecream:
Not at all, don't worry :)

PR #371, opened 2019-02-22
Fix sept-secondary build when using prebuild sept-secondary.enc

igoticecream:
Hello,

This PR fixes the `make dist` after a fresh clone or clean when using the prebuild sept-secondary.enc ($SEPT_ENC_PATH)

Currently failing because sept-secondary.bin is not found

### DIFF
--- a/sept/sept-secondary/Makefile
+++ b/sept/sept-secondary/Makefile
@@ -124,6 +124,7 @@ ifeq ($(strip $(SEPT_ENC_PATH)),)
 	@[ -d $@ ] || mkdir -p $@
 	@$(MAKE) --no-print-directory -C $(BUILD) -f $(CURDIR)/Makefile
 else
+	@touch $(TOPDIR)/sept-secondary.bin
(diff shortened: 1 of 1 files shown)


urherenow:
Sorry for commenting now since this is closed more than a year ago, but I'm confused. This PR makes a 0kb file, while the release comes with a 107,136byte file. Is the file not needed? If so, why is it included in the release .zip? If it is needed/used, how is it not a bug in the build process?

SciresM:
Building sept for real requires keys that can't be dumped publicly.

When building on my machine, it produces the bin file in the release, which is encrypted to make the .enc files. Only the .enc files are actually used.

People who don't have keys pass an environment variable to use the precompiled encrypted binaries, this PR makes the build process work when doing that. Since .enc is passed via variable and not built, the .bin files don't matter and being 0kb isn't a problem.

The bin files are included by make dist so that a) people can load sept in IDA to see I'm not pulling any funny business, and b) whenever keys are public people can decrypt the associated .enc files to make sure they match.

urherenow:
Gotcha. Thanks for the info!

PR #376, opened 2019-02-23
Makefile tweaks

thedax:
-Have fusee-secondary's makefile clean sept-secondary as required.
-Use $(TARGET) variable for sept-secondary, and clean the encrypted output as required.

Currently make clean doesn't clean sept (either stage), and when it does, it leaves the encrypted second stage binary output. I also changed the hard-coded names to be more flexible if the folder name should change in the future.

### DIFF
--- a/fusee/fusee-secondary/Makefile
+++ b/fusee/fusee-secondary/Makefile
@@ -159,6 +159,7 @@ clean:
 	@$(MAKE) -C $(AMS)/exosphere clean
 	@$(MAKE) -C $(AMS)/thermosphere clean
 	@$(MAKE) -C $(AMS)/stratosphere clean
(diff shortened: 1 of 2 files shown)


SciresM:
Sorry, man those buttons need changing.

Anyway -- I don't want to keep merging a bunch of makefile PRs. Do you think you'll be making another one soon?

thedax:
It's cool. And no, I can't think of anything else these need for now. Sorry for the repeated PRs.

SciresM:
@thedax Repeated PRs aren't really a problem, I just figured that if more changes were coming I'd wait a few days before merging.

thedax:
Ah, gotcha. No, this should be the last one like I said in #cfw-dev, assuming I didn't break anything (but I tested fairly thoroughly).

PR #378, opened 2019-02-24
I added in a prebuilt sept-secondary.bin and I added in the SEPT_ENC_PATH variable

piegit:


aveao:
Just my 2c:
Storing binaries on git is a bad idea.

piegit:
Ok sorry

PR #381, opened 2019-02-28
Add logo credit.

sirocyl:


tumGER:
Maybe also add credits for the sept logo then - "||Extra||#2627" seems to be the creator of that.

SciresM:
Art credit stuffs are all encapsulated by `Everyone who submitted entries for Atmosphère's splash design contest.` -- given prize for winning that contest was explicit thanks in README (and since we don't thank the <several> people who also contributed design elements that made it in over the course of the contest), entry should remain the same.

PR #383, opened 2019-03-04
Allow multiple override keys

piepie62:
This was tested by copying the changes over to before all dmnt commits. I can confirm that it works when tested in that way.

### DIFF
--- a/stratosphere/ams_mitm/source/utils.cpp
+++ b/stratosphere/ams_mitm/source/utils.cpp
@@ -487,7 +487,9 @@ bool Utils::HasOverrideButton(u64 tid) {
 }
 
 static OverrideKey ParseOverrideKey(const char *value) {
-    OverrideKey cfg;
+    OverrideKey cfg = {
+        .key_combination = 0
+    };
     
     /* Parse on by default. */
     if (value[0] == '!') {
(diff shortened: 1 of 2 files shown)


SciresM:
Not sure this syntax is desirable -- also, not sure that we prefer an OR of keys rather than an AND....

Needs thought.

piepie62:
Don't quite see how an AND would work, to be quite honest

SciresM:
@piepie62 "require L+R" vs "require L or R", since you're currently doing an OR.

piepie62:
Oh, I thought you meant in the setting code. Yeah, that's a totally valid option, and I could do it pretty easily, if you want

SciresM:
@piepie62 syntax is important -- again, needs thought. Also PR would need to fix in fs.mitm, too, and not just loader.

piepie62:
I thought about doing that, but because it's a fairly minor edit, I decided to PR what I had and get your opinion on it

piepie62:
Oops, sorry, I didn't realize that fs.mitm used the same code. I'll fix that now.

piepie62:
What do you mean by "syntax is important", by the way? If I had more information I might be able to address that issue

piepie62:
Closing this because I don't really have any interest in it anymore. Don't really see a use case in the foreseeable future, and I only really started it originally because I figured it would be possible.

PR #385, opened 2019-03-05
Support for expanded NAND by roblâbla

rashevskyv:
Need for starting with extended NAND chip

### DIFF
--- a/fusee/fusee-secondary/src/nxfs.c
+++ b/fusee/fusee-secondary/src/nxfs.c
@@ -316,7 +316,7 @@ int nxfs_mount_all(void) {
     model = g_mmc_devpart_template;
     model.device_struct = &g_emmc_user_mmcpart;
     model.start_sector = 0;
-    model.num_sectors = (32ull << 30) / model.sector_size;
+    model.num_sectors = (256ull << 30) / model.sector_size;
     
     rc = rawdev_mount_device("rawnand", &model, false);
     


SciresM:
Does this break non-expanded NAND?

rashevskyv:
No issues with both of my consoles. Works great with original 32Gb NAND chip

CTCaer:
why hardcode it and not set it via EXT_CSD_SEC_CNT?

rashevskyv:
> 
> 
> why hardcode it and not set it via EXT_CSD_SEC_CNT?

I do not skilled enough to do it myself, and it was hardcoded by default, but for 32Gb NAND. I had a problem with starting extended NAND on Atmosphere, roblâbla in Discord gave this quick fix for me.

shchmue:
```diff
- model.num_sectors = (256ull << 30) / model.sector_size;
+ model.num_sectors = g_emmc_user_mmcpart.device->ext_csd.sectors;```

shchmue:
oh. oops. the sdmmc code here doesn't init `sectors` for some reason so it'll need to be
`model.num_sectors = *(uint32_t*)g_emmc_user_mmcpart.device->ext_csd.raw_sectors[0];`
or do the math with all four bitshifted bytes in case i did that cast wrong 🤦‍♀️

shchmue:
with a better understanding of the code path, i'll need to think more. fixing this would also probably fix
https://github.com/Atmosphere-NX/Atmosphere/blob/master/fusee/fusee-secondary/src/nxfs.c#L268

PR #389, opened 2019-03-08
fs_mitm: Fix mismatched new[] / delete

leoetlino:
Using delete instead of delete[] on a pointer given by new[] is
undefined behaviour.

### DIFF
--- a/stratosphere/ams_mitm/source/fs_mitm/fsmitm_romfsbuild.cpp
+++ b/stratosphere/ams_mitm/source/fs_mitm/fsmitm_romfsbuild.cpp
@@ -49,7 +49,7 @@ void RomFSBuildContext::VisitDirectory(FsFileSystem *filesys, RomFSBuildDirector
             strcat(child->path + parent->path_len, this->dir_entry.name);
                                     
             if (!this->AddDirectory(parent, child, NULL)) {
-                delete child->path;
+                delete[] child->path;
                 delete child;
             } else {
                 child_dirs.push_back(child);
@@ -72,7 +72,7 @@ void RomFSBuildContext::VisitDirectory(FsFileSystem *filesys, RomFSBuildDirector
             child->size = this->dir_entry.fileSize;
                                     
             if (!this->AddFile(parent, child)) {
-                delete child->path;
+                delete[] child->path;
                 delete child;
             }
         } else {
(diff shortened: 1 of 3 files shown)


SciresM:
If this is really a problem, then this is a problem in a lot of places. Especially in fs.mitm involving romfs paths, which are allocated as new char[]...

Could you make the PR fix every instance of this?

m4xw:
@SciresM this should be fixed.
Should probably fix them step by step if it's deeper in the codebase.

SciresM:
@m4xw I'd rather just fix it in every place all at once, if it's really a problem. fwiw, it seems fine in actual compiled output (IDA shows that it's doing what I'd expect)...but if it's UB then I don't want to do it incrementally. Excise the whole problem.

leoetlino:
Sure, here you go.

SciresM:
Delete[]-ing on memory_source seems incorrect.

For example, we can see [here](https://github.com/Atmosphere-NX/Atmosphere/blob/master/stratosphere/ams_mitm/source/fs_mitm/fsmitm_romfsbuild.cpp#L249) that the RomfsHeader is allocated with new.

However, we can see [here](https://github.com/Atmosphere-NX/Atmosphere/blob/master/stratosphere/ams_mitm/source/fs_mitm/fsmitm_romfsbuild.cpp#L250) that metadata is allocated with new[].

I think we may actually need to switch to malloc/free instead of new/delete, if this is a problem? Man this is dumb.

leoetlino:
Oh sorry, missed the RomfsHeader case...

Yeah hmm it's kind of annoying that this is technically UB. I'll see if there's a nice way to fix this. If not then I'll change it to use malloc/free.

TuxSH:
In the second example, couldn't std::array be used?

leoetlino:
I've replaced new/new[]/delete with malloc/free for memory sources.

I don't think std::array can be used as the size isn't known at compile time.

stratosphere/ams_mitm/source/fs_mitm/fsmitm_romfsbuild.cpp
```diff
@@ -397,13 +399,11 @@ void RomFSBuildContext::Build(std::vector<RomFSSourceInfo> *out_infos) {
     header->dir_table_ofs = header->dir_hash_table_ofs + header->dir_hash_table_size;
     header->file_hash_table_ofs = header->dir_table_ofs + header->dir_table_size;
     header->file_table_ofs = header->file_hash_table_ofs + header->file_hash_table_size;
-    
-    const size_t metadata_size = this->dir_hash_table_size + this->dir_table_size + this->file_hash_table_size + this->file_table_size;
```
SciresM: Why remove this?

SciresM: Oh, you moved it up. Never mind.

leoetlino: I didn't, just moved it up there since the size is needed for malloc anyway

PR #497, opened 2019-04-04
fatal_font: Add missing va_end() calls to PrintFormatLine() and PrintFormat()

lioncash:
Fairly self-explanatory. Ensures va_start() is balanced with va_end()


SciresM:
Thanks!

PR #498, opened 2019-04-05
fs_mitm/fs_dir_utils: Amend logic error within IterateDirectoryRecursivelyInternal

lioncash:
This likely intended to get the size of the string and then subtract the null terminator byte. Instead, this would always result in a sizeof of 8.


SciresM:
Thanks!

Issue #518, opened 2019-04-22
Fatal auto reboot after countdown


fennectech:
#### Overview:
Simple auto reboot timer in fatal.

#### Smaller Details: 
 allow fatal to automaticly reboot 
also for those who want to use the information a fatal providing  a "norebootonfatal.flag" in the atmosphere directory or specify it via config
#### Nature of Request:  
- Change

### What component do you feel this would best fit within?
- [Stratosphère](https://github.com/Atmosphere-NX/Atmosphere#components)</br>
    - Custom system modules.</br>

### Why would this feature be useful?
It would allow users of docked switches not to have to get up to reboot a fataled switch.

HookedBehemoth:
[ Lines between [ ] (square brackets) are to be removed before posting. ]

fennectech:
That better?

HookedBehemoth:
Yes

SciresM:
I don't see why this would be desirable. Hit a button...?

fennectech:
if the console is docked and im sitting back on the couch i'd rather wait a timeout than get up to hit the power button

SciresM:
Implement yourself, make it not the default, and use the setting "atmosphere!fatal_auto_reboot_interval", expressed in milliseconds.

I will accept your PR.

fennectech:
This aint bash      ill see if i can make it but i doubt i will be able to.

PR #519, opened 2019-04-22
Implement Auto Reboot Timer (#518) 

SunTheCourier:
* Config (placed in system_settings under fatal_auto_reboot_interval) takes a u64 which are milliseconds

### DIFF
--- a/common/defaults/system_settings.ini
+++ b/common/defaults/system_settings.ini
@@ -6,6 +6,9 @@ upload_enabled = u8!0x0
 usb30_force_enabled = u8!0x0
 ; Atmosphere custom settings
 [atmosphere]
+; Reboot from fatal automatically after 5 seconds (in milliseconds)
+; If field is disabled fatal waits for an input indefinitely
+; fatal_auto_reboot_interval = u64!‪5000‬
 ; Make the power menu's "reboot" button reboot to payload.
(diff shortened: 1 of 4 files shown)


SciresM:
Please convert to milliseconds, and make it u64 instead of s64 -- thank you!

SciresM:
Also, please use a TimeoutHelper -- you can see an example of that in use for quest units in the same function -- no need for it to be separate logic.

SunTheCourier:
Done :)

SciresM:
Looks good. Could you also do `config->is_auto_reboot_enabled &= (config->fatal_auto_reboot_interval != 0);`?

This would allow setting a default value of 0 (disabled), which would be helpful from a present-in-system-settings.ini-as-an-example perspective.

SunTheCourier:
Should 0 not be allowed as an actual value? (no delay with reboot on fatal)

SciresM:
Yes, 0 should not be allowed as an actual value.

SunTheCourier:
Did i get `config->is_auto_reboot_enabled &= (config->fatal_auto_reboot_interval != 0);` correctly?

SciresM:
lgtm. Can you test on hardware? If you confirm it works, I will merge.

SunTheCourier:
Manual button input and automatic reboot seem to function fine

SciresM:
Great, thanks again!

PR #520, opened 2019-04-22
Reimplement the ro sysmodule.

SciresM:
This PR reimplements the `ro` sysmodule (`ldr:ro`, `ro:dmnt`).

It also makes Atmosphere provide ldr:ro/ro:dmnt on all firmware versions -- ro sysmodule was introduced on 3.0.0, but we will host it on every firmware version.

This allows moving all nro related code out of `loader`, which is nice, and also allows adding support for applying IPS patches to NROs. We support those the same way we support exefs patches, via /atmosphere/nro_patches/<*>/<build_id>.ips.

Making PR now because stuff is fully functional (tested on 1.0.0, 8.0.0) -- will merge in a few days following more testing + maybe cleanup.


SciresM:
Meh, gonna merge this now since otherwise merge conflicts will get pretty bad.

PR #521, opened 2019-04-22
Uncomment fatal_auto_reboot_interval and set to 0

SunTheCourier:
* Would work better for homebrew apps such as Kosmos Toolbox

### DIFF
--- a/common/defaults/system_settings.ini
+++ b/common/defaults/system_settings.ini
@@ -6,9 +6,10 @@ upload_enabled = u8!0x0
 usb30_force_enabled = u8!0x0
 ; Atmosphere custom settings
 [atmosphere]
-; Reboot from fatal automatically after 5 seconds (in milliseconds)
+; Reboot from fatal automatically (in milliseconds)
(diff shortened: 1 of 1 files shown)


HookedBehemoth:
Then they should do it themself in their release.
The fatal screen is there for something and should not be ignored by default.

cole-h:
Setting it to 0 functions the same as having it commented. My opinion doesn't count for much, but I don't think this change is controversial. 

https://github.com/Atmosphere-NX/Atmosphere/blob/153e97a80faf8f92aa80dcc10d09a3c0eb50243e/stratosphere/fatal/source/fatal_config.cpp#L88

HookedBehemoth:
Sorry, I overlooked that part.

SunTheCourier:
As i said it's so homebrew like Kosmos Toolbox can modify the value properly

SciresM:
I'll do this when I add ro!ease_nro_restriction default. Not relevant until next release anyway.

PR #532, opened 2019-04-30
Implement PRODINFO blanking and redirection

### LINKED ISSUE #165: Option to destroy/mask client-specific keys, certificate and data while using Atmosphere.
This is specifically to prevent one's console from being banned due to accidental communication with Nintendo or their servers while under an environment which would violate sanity checks, present or future.

Nintendo servers will see nonsense or zero data for units with homebrew and CFW in use, rather than a client's serial, deviceid, or other per-console unique information, and refuse access to connect.

For implementation, the original, unmodified data can remain as-is in place, where client data masking is only emulated under TZ or ssl/spl module, or be backed up to a safe, persistent location and destroyed in place on the flash in PRODINFO.

Regardless of implementation, this process _must_ be reversible by the end-user.

ThatNerdyPikachu:
This PR introduces a way for the user to "zero out" their console's PRODINFO partition, wiping console identification information, by means of redirecting the partition to a file stored on the SD card, instead of directly modifying the partition and risking the console. This is controlled by a global flag, ``blank_prodinfo``, which will redirect the partition if it's present, and return the normal partition if it is not.

This closes #165.

### DIFF
--- a/stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp
+++ b/stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp
@@ -32,6 +32,7 @@
 #include "fs_save_utils.hpp"
 #include "fs_subdirectory_filesystem.hpp"
 #include "fs_directory_savedata_filesystem.hpp"
+#include "fs_file_storage.hpp"
 
 #include "../debug.hpp"
 
@@ -264,9 +265,15 @@ Result FsMitmService::OpenBisStorage(Out<std::shared_ptr<IStorageInterface>> out
             if (bis_partition_id == BisStorageId_Boot0) {
                 storage = std::make_shared<IStorageInterface>(new Boot0Storage(bis_storage, this->title_id));
             } else if (bis_partition_id == BisStorageId_Prodinfo) {
+                bool has_blank_cal0_flag = true; // See PR details
+
                 /* PRODINFO should *never* be writable. */
                 if (is_sysmodule || has_cal0_read_flag) {
                     storage = std::make_shared<IStorageInterface>(new ROProxyStorage(bis_storage));
+                } else if (has_blank_cal0_flag) {
+                    FsFile file = Utils::GetBlankProdinfoFileHandle();
+                    
+                    storage = std::make_shared<IStorageInterface>(new FileStorage(std::make_shared<ProxyFile>(&file)));
                 } else {
                     /* Do not allow non-sysmodules to read *or* write CAL0. */
                     fsStorageClose(&bis_storage);
--- a/stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp
+++ b/stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp
@@ -269,12 +269,12 @@ Result FsMitmService::OpenBisStorage(Out<std::shared_ptr<IStorageInterface>> out
                 if (has_blank_cal0_flag) {
                     FsFile file;
 
-                    rc = Utils::OpenBlankProdinfoFile(&file);
+                    rc = Utils::OpenBlankProdInfoFile(&file);
                     if (R_FAILED(rc)) {
                         return rc;
                     }
                     
-                    storage = std::make_shared<IStorageInterface>(new FileStorage(new ProxyFile(&file))); /* Should we make this read-only? */
+                    storage = std::make_shared<IStorageInterface>(new FileStorage(new ProxyFile(&file)));
                 } else if (is_sysmodule || has_cal0_read_flag) {
                     /* PRODINFO should *never* be writable. */
                     storage = std::make_shared<IStorageInterface>(new ROProxyStorage(bis_storage));
--- a/stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp
+++ b/stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp
@@ -262,18 +262,22 @@ Result FsMitmService::OpenBisStorage(Out<std::shared_ptr<IStorageInterface>> out
             const bool is_sysmodule = TitleIdIsSystem(this->title_id);
             const bool has_bis_write_flag = Utils::HasFlag(this->title_id, "bis_write");
             const bool has_cal0_read_flag = Utils::HasFlag(this->title_id, "cal_read");
+            const bool has_blank_cal0_flag = Utils::HasGlobalFlag("blank_prodinfo");
             if (bis_partition_id == BisStorageId_Boot0) {
                 storage = std::make_shared<IStorageInterface>(new Boot0Storage(bis_storage, this->title_id));
             } else if (bis_partition_id == BisStorageId_Prodinfo) {
-                bool has_blank_cal0_flag = true; // See PR details
+                if (has_blank_cal0_flag) {
+                    FsFile file;
 
-                /* PRODINFO should *never* be writable. */
-                if (is_sysmodule || has_cal0_read_flag) {
-                    storage = std::make_shared<IStorageInterface>(new ROProxyStorage(bis_storage));
-                } else if (has_blank_cal0_flag) {
-                    FsFile file = Utils::GetBlankProdinfoFileHandle();
+                    rc = Utils::OpenBlankProdinfoFile(&file);
+                    if (R_FAILED(rc)) {
+                        return rc;
+                    }
                     
-                    storage = std::make_shared<IStorageInterface>(new FileStorage(std::make_shared<ProxyFile>(&file)));
+                    storage = std::make_shared<IStorageInterface>(new FileStorage(new ProxyFile(&file))); /* Should we make this read-only? */
+                } else if (is_sysmodule || has_cal0_read_flag) {
+                    /* PRODINFO should *never* be writable. */
+                    storage = std::make_shared<IStorageInterface>(new ROProxyStorage(bis_storage));
                 } else {
                     /* Do not allow non-sysmodules to read *or* write CAL0. */
                     fsStorageClose(&bis_storage);
--- a/stratosphere/ams_mitm/source/utils.cpp
+++ b/stratosphere/ams_mitm/source/utils.cpp
@@ -673,7 +673,7 @@ void Utils::RebootToFatalError(AtmosphereFatalErrorContext *ctx) {
 void Utils::CreateBlankProdInfoIfNeeded() {
     Result rc;
 
-    rc = fsFsOpenFile(&g_sd_filesystem, "/atmosphere/automatic_backups/prodinfo_blank.bin", FS_OPEN_READ, &g_blank_prodinfo_file);
+    rc = fsFsOpenFile(&g_sd_filesystem, "/atmosphere/automatic_backups/prodinfo_blank.bin", FS_OPEN_READ | FS_OPEN_WRITE, &g_blank_prodinfo_file);
     if (R_SUCCEEDED(rc)) {
         return;
     }
@@ -723,7 +723,7 @@ void Utils::CreateBlankProdInfoIfNeeded() {
         memcpy(&p[hash_offset], hash, 0x20);
     }
 
-    rc = fsFsOpenFile(&g_sd_filesystem, "/atmosphere/automatic_backups/prodinfo_blank.bin", FS_OPEN_WRITE, &g_blank_prodinfo_file);
+    rc = fsFsOpenFile(&g_sd_filesystem, "/atmosphere/automatic_backups/prodinfo_blank.bin", FS_OPEN_READ | FS_OPEN_WRITE, &g_blank_prodinfo_file);
     if (R_FAILED(rc)) {
         std::abort();
     }
--- a/stratosphere/ams_mitm/source/utils.cpp
+++ b/stratosphere/ams_mitm/source/utils.cpp
@@ -708,7 +708,7 @@ void Utils::CreateBlankProdinfoIfNeeded() {
 
         u8 data[data_size];
 
-        memcpy(&data, p ? g_cal0_backup[data_offset] : g_cal0_storage_backup[data_offset], data_size);
+        memcpy(&data, p ? &g_cal0_backup[data_offset] : &g_cal0_storage_backup[data_offset], data_size);
 
         u8 hash[0x20];
 
--- a/stratosphere/ams_mitm/source/utils.cpp
+++ b/stratosphere/ams_mitm/source/utils.cpp
@@ -65,6 +65,8 @@ static FsFile g_cal0_file = {0};
 static u8 g_cal0_storage_backup[ProdinfoSize];
 static u8 g_cal0_backup[ProdinfoSize];
 
+static FsFile g_blank_prodinfo_file = {0};
+
 static bool IsHexadecimal(const char *str) {
     while (*str) {
         if (isxdigit(*str)) {
@@ -139,7 +141,7 @@ void Utils::InitializeThreadFunc(void *args) {
                 fsFileFlush(&g_cal0_file);
             }
             
-            Utils::CreateBlankProdinfoIfNeeded();
+            Utils::CreateBlankProdInfoIfNeeded();
 
             /* NOTE: g_cal0_file is intentionally not closed here. This prevents any other process from opening it. */
             memset(g_cal0_storage_backup, 0, sizeof(g_cal0_storage_backup));
@@ -668,24 +670,23 @@ void Utils::RebootToFatalError(AtmosphereFatalErrorContext *ctx) {
     BpcRebootManager::RebootForFatalError(ctx);
 }
 
-void Utils::CreateBlankProdinfoIfNeeded() {
-    const bool p = Utils::GetCAL0BackupBufferToBlank();
-    
+void Utils::CreateBlankProdInfoIfNeeded() {
     Result rc;
-    FsFile file;
 
-    rc = fsFsOpenFile(&g_sd_filesystem, "/atmosphere/automatic_backups/prodinfo_blank.bin", FS_OPEN_READ, &file);
+    rc = fsFsOpenFile(&g_sd_filesystem, "/atmosphere/automatic_backups/prodinfo_blank.bin", FS_OPEN_READ, &g_blank_prodinfo_file);
     if (R_SUCCEEDED(rc)) {
-        fsFileClose(&file);
-        
         return;
     }
+
+    fsFsCreateFile(&g_sd_filesystem, "/atmosphere/automatic_backups/prodinfo_blank.bin", ProdinfoSize, 0);
+
+    u8* p = Utils::GetCal0BufferToBlank();
 
     // big brain: modify buffer inline
 
     const char serial[] = "XAW00000000000";
 
-    memcpy(p ? &g_cal0_backup[0x250] : &g_cal0_storage_backup[0x250], serial, sizeof(serial) - 1);
+    memcpy(&p[0x250], serial, sizeof(serial) - 1);
 
     static constexpr u64 erase_offsets[7] = {0xAE0, 0x3AE0, 0x35E1, 0x36E1, 0x2B0, 0x3D70, 0x3FC0};
     static constexpr u64 erase_sizes[7] = {0x800, 0x130, 0x6, 0x6, 0x180, 0x240, 0x240};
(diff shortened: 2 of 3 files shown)


stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp
```diff
@@ -264,9 +265,15 @@ Result FsMitmService::OpenBisStorage(Out<std::shared_ptr<IStorageInterface>> out
             if (bis_partition_id == BisStorageId_Boot0) {
                 storage = std::make_shared<IStorageInterface>(new Boot0Storage(bis_storage, this->title_id));
             } else if (bis_partition_id == BisStorageId_Prodinfo) {
+                bool has_blank_cal0_flag = true; // See PR details
```
SciresM: Should always default to false.

stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp
```diff
@@ -264,9 +265,15 @@ Result FsMitmService::OpenBisStorage(Out<std::shared_ptr<IStorageInterface>> out
...
+
                 /* PRODINFO should *never* be writable. */
                 if (is_sysmodule || has_cal0_read_flag) {
                     storage = std::make_shared<IStorageInterface>(new ROProxyStorage(bis_storage));
+                } else if (has_blank_cal0_flag) {
```
SciresM: Wrong if ordering -- this will just result in this never doing the thing you want.

stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp
```diff
@@ -264,9 +265,15 @@ Result FsMitmService::OpenBisStorage(Out<std::shared_ptr<IStorageInterface>> out
...
+                    FsFile file = Utils::GetBlankProdinfoFileHandle();
+                    
+                    storage = std::make_shared<IStorageInterface>(new FileStorage(std::make_shared<ProxyFile>(&file)));
```
SciresM: This probably needs special handling to not leak domain IDs.

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,88 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
 void Utils::RebootToFatalError(AtmosphereFatalErrorContext *ctx) {
     BpcRebootManager::RebootForFatalError(ctx);
 }
+
+void Utils::CreateBlankProdinfoIfNeeded() {
+    /* if (!Utils::HasGlobalFlag("blank_prodinfo")) {
```
SciresM: This will never work because the SD card is not initialized at the time you're doing this. But you want to do it unconditionally anyway, so just remove it.

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,88 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+
+    Result rc;
+
+    u8* buffer = g_cal0_backup;
+
+    // big brain: modify buffer inline
+
+    const char* serial = "XAW00000000000"; // TODO: Make this configurable?
+
+    memcpy(&buffer[0x250], serial, strlen(serial));
```
SciresM: memcpy(..., "XAW..", ...);

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,88 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+
+    const u64 erasePoints[14] = {0xAE0, 0x800, 0x3AE0, 0x130, 0x35E1, 0x6, 0x36E1, 0x6, 0x2B0, 0x180, 0x3D70, 0x240, 0x3FC0, 0x240}; // offset, size
```
SciresM: Please do not store buffers that alternate different values. It is really bad style.

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,88 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+
+    for (u64 i = 0; i < 14; i = i + 2) {
```
SciresM: size_t

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,88 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+        u64 offset = erasePoints[i];
+        u64 size = erasePoints[i + 1];
+
+        memset(&buffer[offset], 0, size);
+    }
+
+    const u64 hashPoints[6] = {0x12E0, 0xAE0, buffer[0xAD0], 0x20, 0x40, buffer[0x8]}; // hash offset, data offset, data size
+
+    for (u64 i = 0; i < 6; i = i + 3) {
```
SciresM: size_t

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,88 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+
+        sha256ContextUpdate(&ctx, data, dataSize);
+
+        sha256ContextGetHash(&ctx, &hash);
+
+        memcpy(&buffer[hashOffset], hash, 0x20);
+    }
+
+    rc = fsFsCreateFile(&g_sd_filesystem, "/atmosphere/automatic_backups/prodinfo_blank.bin", ProdinfoSize, 0);
+    if (R_FAILED(rc)) {
```
SciresM: This will cause abort on bootup if the file exists, no need to check result.

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,88 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+        std::abort();
+    }
+
+    rc = fsFileFlush(&file);
+    if (R_FAILED(rc)) {
+        std::abort();
+    }
+
+    fsFileClose(&file);
```
SciresM: You should persist a read handle to this, so that the file may not be opened by other processes. See what we do for the normal PRODINFO backup. Instead of reopening, you can also just do a duplicate session.

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,88 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+}
+
+FsFile Utils::GetBlankProdinfoFileHandle() {
```
SciresM: Result Utils::OpenBlankProdInfoFile(FsFile *out)

stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp
```diff
@@ -264,9 +265,15 @@ Result FsMitmService::OpenBisStorage(Out<std::shared_ptr<IStorageInterface>> out
             if (bis_partition_id == BisStorageId_Boot0) {
                 storage = std::make_shared<IStorageInterface>(new Boot0Storage(bis_storage, this->title_id));
             } else if (bis_partition_id == BisStorageId_Prodinfo) {
+                bool has_blank_cal0_flag = true; // See PR details
+
                 /* PRODINFO should *never* be writable. */
                 if (is_sysmodule || has_cal0_read_flag) {
                     storage = std::make_shared<IStorageInterface>(new ROProxyStorage(bis_storage));
+                } else if (has_blank_cal0_flag) {
+                    FsFile file = Utils::GetBlankProdinfoFileHandle();
```
SciresM: Error checking.

stratosphere/ams_mitm/source/utils.cpp
ThatNerdyPikachu: What do you mean by alternate different values?

stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp
ThatNerdyPikachu: But storage could just be reassinged twice, that was my reasoning..?

SciresM: this is in an else if, this ordering is wrong.

ThatNerdyPikachu: oh, was thinking it was just another if block. oops

stratosphere/ams_mitm/source/utils.cpp
SciresM: You're storing two conceptually different things (offset, value) in a single buffer.

Also can mark this static constexpr.

ThatNerdyPikachu: Why constexpr over const?

SciresM: const will actually initialize the buffer on stack.

marking it static will move it into .rodata

static constexpr is preferable over static const because it lets the compiler know it's knowable at compiltime.

Adubbz:
Mentioned this on Discord but will also put this here for record keeping. It needs to be confirmed that nothing from prodinfo is cached and can potentially be compared with the current values and subsequently be included in telemetry. This was a potential issue with vertigo's implementation too which I don't believe was looked into.

stratosphere/ams_mitm/source/utils.cpp
SciresM: Actually:

const char serial[] = "XAW...";
memcpy..., serial, sizeof(serial));

stratosphere/ams_mitm/source/utils.cpp
ThatNerdyPikachu: So just make something like ``g_blank_prodinfo_file`` and return that in the OpenFile function?

stratosphere/ams_mitm/source/utils.cpp
ThatNerdyPikachu: So just make something like ``g_blank_prodinfo_file`` and return that in the OpenFile function?

ThatNerdyPikachu:
@Adubbz Do you have any suspects of processes that would cache this data, as a starting point to look?

stratosphere/ams_mitm/source/utils.cpp
ThatNerdyPikachu: Why do the fsFsCreateFile calls work then?

SciresM:
```
bool Utils::HasGlobalFlag(const char *flag) {
    if (IsSdInitialized()) {
```

```
bool Utils::IsSdInitialized() {
    return g_has_initialized;
}```

`g_has_initialized` is not set until the end of the function that calls the code you're writing.

File: stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp (hunk 3 above)
Comment (ThatNerdyPikachu) [reply stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp]: How would I go about that?

**Comment by Adubbz** [comment]
Anything with access to set:cal (as well as settings itself) should be checked. Here's a list:
- am
- audio
- bluetooth
- es
- hid
- nim
- nvnflinger
- pctl
- pcv
- ppc
- ptm
- settings
- ssl
- wlan

**Comment by ThatNerdyPikachu** [comment]
With a new commit comes stupidity, broken things, and more.

# Fixes
- Virtually everything on @SciresM's list, except for the 1 currently marked as unresolved, which I still do not know how to fix.
- Some other things that were stupid and should of never happened.

# Bugs
- ``g_cal0_storage_backup`` is always corrupted, somehow...
- When the blank PRODINFO file is present on the SD card, console fails to boot (``std::abort()``). **I still cannot find the cause of this, after half an hour of debugging**
- Console ``std::abort()``'s while trying to read PRODINFO, even with the flag not present. Probably the fault of the file opener function.
- Bad code

File: stratosphere/ams_mitm/source/fs_mitm/fsmitm_service.cpp [hunk 13]
```diff
@@ -261,11 +262,21 @@ Result FsMitmService::OpenBisStorage(Out<std::shared_ptr<IStorageInterface>> out
...
-                if (is_sysmodule || has_cal0_read_flag) {
+                if (has_blank_cal0_flag) {
+                    FsFile file;
+
+                    rc = Utils::OpenBlankProdinfoFile(&file);
+                    if (R_FAILED(rc)) {
+                        return rc;
+                    }
+                    
+                    storage = std::make_shared<IStorageInterface>(new FileStorage(new ProxyFile(&file))); /* Should we make this read-only? */
```
SciresM: It's already read-only because it's opened FS_OPEN_READ.

Also, this still leaks object IDs, but I'll just fix that when it's merged as explaining why might be complicated.

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,91 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
 void Utils::RebootToFatalError(AtmosphereFatalErrorContext *ctx) {
     BpcRebootManager::RebootForFatalError(ctx);
 }
+
+void Utils::CreateBlankProdinfoIfNeeded() {
```
SciresM: ProdInfo

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,91 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+    const bool p = Utils::GetCAL0BackupBufferToBlank();
```
SciresM: Have this actually return a buffer instead of a bool. This will remove ugly p ? x : y from the rest of code.

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,91 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+    
+    Result rc;
+    FsFile file;
+
+    rc = fsFsOpenFile(&g_sd_filesystem, "/atmosphere/automatic_backups/prodinfo_blank.bin", FS_OPEN_READ, &file);
+    if (R_SUCCEEDED(rc)) {
+        fsFileClose(&file);
+
```
SciresM: Extraneous newline

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,91 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+        Sha256Context ctx;
+        sha256ContextCreate(&ctx);
+
+        sha256ContextUpdate(&ctx, data, data_size);
+
+        sha256ContextGetHash(&ctx, &hash);
+
+        memcpy(p ? &g_cal0_backup[hash_offset] : &g_cal0_storage_backup[hash_offset], hash, 0x20);
+    }
+
```
SciresM: Where is the fsCreateFile? You should unconditionally try to create it without checking the error code, prior to the open.

stratosphere/ams_mitm/source/utils.cpp
SciresM: This still wants a read handle persisted.

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,91 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+    rc = fsFileFlush(&file);
+    if (R_FAILED(rc)) {
+        std::abort();
+    }
+
+    fsFileClose(&file);
+}
+
+Result Utils::OpenBlankProdinfoFile(FsFile *out) {
+    return fsFsOpenFile(&g_sd_filesystem, "/atmosphere/automatic_backups/prodinfo_blank.bin", FS_OPEN_READ, out);
```
SciresM: Duplicate persisted read handle instead of opening new reference.

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,91 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+}
+
+bool Utils::GetCAL0BackupBufferToBlank() { // true = g_cal0_backup, false = g_cal0_storage_backup
+    bool is_cal0_valid = true;
+    is_cal0_valid &= memcmp(g_cal0_backup, "CAL0", 4) == 0;
+
```
SciresM: Do actual CAL0 validation, please see code elsewhere in Utils.cpp for doing this. Better: `if (IsCal0Valid(g_cal0_backup)) { /* ... */ else if (IsCal0Valid(g_cal0_storage_backup)) { /* ... */ } else { std::abort(); }`

stratosphere/ams_mitm/source/utils.cpp
```diff
@@ -665,3 +667,91 @@ Result Utils::GetSettingsItemBooleanValue(const char *name, const char *key, boo
...
+    }
+
+    fsFileClose(&file);
+}
+
+Result Utils::OpenBlankProdinfoFile(FsFile *out) {
+    return fsFsOpenFile(&g_sd_filesystem, "/atmosphere/automatic_backups/prodinfo_blank.bin", FS_OPEN_READ, out);
+}
+
+bool Utils::GetCAL0BackupBufferToBlank() { // true = g_cal0_backup, false = g_cal0_storage_backup
```
SciresM: Cal0

stratosphere/ams_mitm/source/utils.cpp
ThatNerdyPikachu: Heck... did I never readd it?

stratosphere/ams_mitm/source/utils.cpp
ThatNerdyPikachu: https://canary.discordapp.com/channels/269333940928512010/420029476634886144/572926601906356234

stratosphere/ams_mitm/source/utils.cpp
ThatNerdyPikachu: Yeah, took that code from there, actually.

ThatNerdyPikachu:
@SciresM The CAL0 storage backup still seems to be corrupted, I’m going to go about a different approach to fixing it, you can review it then, as it isn’t desirable.

SciresM:
> @SciresM The CAL0 storage backup still seems to be corrupted, I’m going to go about a different approach to fixing it, you can review it then, as it isn’t desirable.

Given that automatic backups work, we can conclude the storage backup is not corrupted.

ThatNerdyPikachu:
@SciresM Writing the storage backup to a file always leads to a garbage file.

ThatNerdyPikachu:
Things still don't work:

- No matter what, while opening/reading from the CAL0 partition, with or without the flag, we get a std::abort().
- ``g_cal0_storage_backup`` reports that it is valid with the method, but on write, produces a garbage file.

@SciresM

clohito:
I support this

SciresM:
https://github.com/Atmosphere-NX/Atmosphere/commit/3bba035b8483faf029341a83bd2ab9baad6a657e

SciresM:
I wish I could report a happier outcome, but a summary of the above, reposting here for visibility:

```
Implement blank prodinfo creation.

This is a complete implementation of what PR #532 seeks to do
(thanks @ThatNerdyPikachu). However, it currently blackscreens.

This is because we can't actually mitm settings, because settings
must be able to complete its initialization before the sd card
can be mounted. Thus we end up with a circular dependency and
the console blackscreens. This problem may yet be solvable, but
it's unclear immediately how this dependency might be solved. In any
event, this serves as a reference to use in the event that a solution
arises.
```

Drakulix:
> This is because we can't actually mitm settings, because settings
must be able to complete its initialization before the sd card
can be mounted. Thus we end up with a circular dependency and
the console blackscreens.

I do not have much knowledge on the horizon interna, but should it not be possible to use a static buffer to mitm the prodinfo partition instead of a file on the sd card? Or is the sd card depedency introduced by something else?

I would really like to see this feature being implemented and the code changes seem simple enough for me to continue the work on it, if this would be a valid way forward.

SciresM:
@Drakulix the static buffer cannot be populated until the SD card is initialized. Static buffer cannot be used for ProdInfo, it must be initialized with console specific data, or else e.g. the battery will be miscalibrated and bad hardware stuff can happen.

I am unsure whether settings will access more partitions than ProdInfo, but in memory could maybe be viable if not.

Drakulix:
>  I am unsure whether settings will access more partitions than ProdInfo, but in memory could maybe be viable if not.

So the idea would be to read console's PRODINFO into a memory buffer, blank out every non-hardware related field and then use that to mitm settings?

SciresM:
yeah, dunno if that would actually work, but basically.

GavinDarkglider:
I would think that once emunand is released we could spoof/blank the data on the emunand. Also another random thought, couldnt you do that in the fusee payload? I know hekate does something similar by putting an arbitrary value in a pointer, and recalling that memory address in coreboot, and checking what is there. That probably isnt viable though, as the system booting would probably overwrite the buffer before it was needed.

PR #546, opened 2019-05-23
Implement PRODINFO blanking and redirection

SciresM:
This is a fully functional + tested implementation of #532.

Note, copying from @Adubbz comment on #532:

Mentioned this on Discord but will also put this here for record keeping. It needs to be confirmed that nothing from prodinfo is cached and can potentially be compared with the current values and subsequently be included in telemetry. This was a potential issue with vertigo's implementation too which I don't believe was looked into.


designgears:
As mentioned on https://mkey.salthax.org/ the device ID can be derived from fuses or obtained from PRODINFO. Not sure if that matters.

SciresM:
Superceded by #913

PR #548, opened 2019-05-25
creport: Add 32 bits stack frames parsing support

Thog:
Also fix FP, SP and LR registers being set wrongly by svcGetDebugThreadParam for 32 bits processes.


SciresM:
Please change stackframe structure to 
```
struct StackFrame {
    union {
        ...
    } frame_64;
    union {
        ...
    } frame_32;
};
```
please.

Other than that, looks good.

Thog:
Done

PR #550, opened 2019-06-04
setsys_settings_items: Minor cleanup to string buffer management

lioncash:
Just tidies up the string management and makes for roughly 3 less discarded copies strings when constructing the combined name and key by using the initial std::string constructed and appending onto it. It also simplifies the text file reading in LoadConfiguration, as c_str() will always terminate it's contained string, so we can simplify the logic used to read from the file a tiny bit.


SciresM:
Thanks! :)

PR #555, opened 2019-06-04
bpcmitm_main: Add missing header guard

lioncash:
Prevents any potential double inclusion issues from occurring. While we're at it, we can also tidy up the header's included files as well, given only the type definitions from <switch.h> are used by it. The corresponding cpp file already includes these headers.


SciresM:
Thanks!

PR #561, opened 2019-06-08
dmnt:cht not loading cheats with long cheat names, even if they are just titles/credit

WerWolv:
The main intention of this PR is to allow loading of all cheats again. Before, if cheat creators added credits or other information to the cheat file and the text got longer than 64 characters, dmnt:cht would refuse to load the cheat. Now if the name gets too long, it simply gets truncated and the parsing carries on.

I also doubled the maximum cheat file size from 2kB to 4kB and the maximum number of opcodes from 256 to 512, mainly because some cheats that that gave the user a unlimited number of multiple items (For example the All Candy x999 cheat for Pokemon Let's Go) are pretty long and caused the whole cheat to fail loading.

Last but not least I fixed a compilation issue with the latest libnx changes (namely fsFileRead and fsFileWrite)

### DIFF
--- a/stratosphere/dmnt/source/dmnt_cheat_manager.cpp
+++ b/stratosphere/dmnt/source/dmnt_cheat_manager.cpp
@@ -321,13 +321,14 @@ bool DmntCheatManager::ParseCheats(const char *s, size_t len) {
             size_t j = i + 1;
             while (s[j] != ']') {
                 j++;
-                if (j >= len || (j - i - 1) >= sizeof(cur_entry->definition.readable_name)) {
+                if (j >= len) {
                     return false;
                 }
             }
 
             /* s[i+1:j] is cheat name. */
-            const size_t cheat_name_len = (j - i - 1);
+            /* Truncate the cheat name if it's too long */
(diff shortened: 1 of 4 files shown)


SciresM:
>Truncation of cheat names

This is actually pretty acceptable/a good idea.

>Doubling cheat size

Worried about memory constraints, since the original sizes were chosen specifically with that in mind. Keep in mind the *whole VM* can only have 0x400 opcodes, and that's not changing without serious thought/discussion.

>fsFileRead/fsFileWrite

Fixed in development branch already, interested in avoiding merge conflicts >.>

WerWolv:
Hm okay, didn't know about that limitation. Anything that can be done about those super long cheats? I haven't seen many so far but they apparently work on Gateway's implementation so people keep making them.
I can revert these two commits though.

SciresM:
'they apparently work on Gateway's implementation so people keep making them." -- Yeah, hrm. Gimme a chew on it, it's because Gateway doesn't statically reserve space for their cheats (since they don't have the VM reinitialization/toggleability-in-process that we do).

SciresM:
Closing since it can't be merged anyway + dmnt refactor can be made to include what we want.

PR #562, opened 2019-06-09
Help compilers with sept clarifications

leo60228:
@clohito (@usi#8810) is having problems making this PR (possibly because his GitHub account is new?), so I made it for him. Please direct criticism to him.


DavidBuchanan314:
@leo60228 It might be that they were banned from the github org...

PR #563, opened 2019-06-12
Fix documentation for sept's python dependencies

ammaraskar:
Just ran into this issue following the docs, apparently installing both together is a bad idea https://pycryptodome.readthedocs.io/en/latest/src/installation.html#installation

> One must avoid having both PyCrypto and PyCryptodome installed at the same time, as they will interfere with each other.


hexkyz:
Confirmed. I was able to trigger this issue in a fresh new environment. PyCryptodome is the best suggestion anyway so, merged.

PR #564, opened 2019-06-14
fix usage of ctype macros

WinterMute:
Passing negative values to the macros in ctype.h other than EOF is undefined behaviour, see caveats section at https://netbsd.gw.com/cgi-bin/man-cgi?ctype++NetBSD-current

### DIFF
--- a/fusee/fusee-primary/src/lib/vsprintf.c
+++ b/fusee/fusee-primary/src/lib/vsprintf.c
@@ -71,7 +71,7 @@ static inline char *pack_hex_byte(char *buf, uint8_t byte)
  */
 static char *skip_spaces(const char *str)
 {
-	while (isspace(*str))
+	while (isspace((unsigned char)*str))
 		++str;
 	return (char *)str;
 }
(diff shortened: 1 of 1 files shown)


misson20000:
Not sure where negative chars would even come from here or how casting them to unsigned is much better than invoking undefined behavior (there's nothing sane to do with them; casting them to unsigned is not a sane thing to do with them), but I'm guessing you mostly just want to silence a compiler warning?

WinterMute:
I'll reproduce the caveats section here for easier reference. This explicitly recommends casting to unsigned char in order to prevent the undefined behaviour.

```
  CAVEATS
     The argument of these functions is of type int, but only a very
     restricted subset of values are actually valid.  The argument must either
     be the value of the macro EOF (which has a negative value), or must be a
     non-negative value within the range representable as unsigned char.
     Passing invalid values leads to undefined behavior.

     Values of type int that were returned by getc(3), fgetc(3), and similar
     functions or macros are already in the correct range, and may be safely
     passed to these ctype functions without any casts.

     Values of type char or signed char must first be cast to unsigned char,
     to ensure that the values are within the correct range.  Casting a nega-
     tive-valued char or signed char directly to int will produce a negative-
     valued int, which will be outside the range of allowed values (unless it
     happens to be equal to EOF, but even that would not give the desired
     result).

     Because the bugs may manifest as silent misbehavior or as crashes only
     when fed input outside the US-ASCII range, the NetBSD implementation of
     the ctype functions is designed to elicit a compiler warning for code
     that passes inputs of type char in order to flag code that may pass nega-
     tive values at runtime that would lead to undefined behavior:
```

PR #584, opened 2019-06-20
sept: use python2 in Makefile

misson20000:
Makes life easier for distros like Arch where `python` defaults to Python 3.


TuxSH:
Is sept_sign not python2 compatible?

DavidBuchanan314:
Yes, it should be. Although, I maybe there's been a regression since I don't believe SciresM tests with python3.

misson20000:
nvm it is, I'd had the `crypto` package installed on python3 instead of `pycryptodome`... kinda bitter that `pycryptodome`'s namespace is `Crypto`.

Issue #560, opened 2019-06-08
[Meta] Refactor to use namespaces


SciresM:
This is per #cfw-development vote.

hac::?

ams::?

sts::?

Namespaces are a good feature and we should use them instead of doing the static class shit we do now.

SciresM:
we did it, reddit

PR #585, opened 2019-06-21
Completely re-write sm.

SciresM:
This rewrites sm to use newer stratosphere coding style/standards.

It's written using namespaces, which we decided to use based on a vote. See #560.

Code lives inside sts::sm::.


stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+        struct ProcessInfo {
+            u64 pid;
+            size_t access_control_size;
+            u8 access_control[AccessControlSizeMax];
+        };
+
+        struct ServiceInfo {
+            ServiceName name;
+            u64 owner_pid;
+            Handle port_h;
```
misson20000: Would be nice to see handles managed via RAII and move semantics

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+                R_TRY(resp->result);
+                should_mitm = resp->should_mitm;
+            }
+
+            /* If we shouldn't mitm, give normal session. */
+            if (!should_mitm) {
+                return svcConnectToPort(out, service_info->port_h);
+            }
+
+            /* Create both handles. If the second fails, close the first to prevent leak. */
```
misson20000: Or just use RAII and let the compiler figure it out :man_shrugging:

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+            *out = 0;
+
+            /* If not mitm'd or mitm service is requesting, get normal session. */
+            if (!IsValidProcessId(service_info->mitm_pid) || service_info->mitm_pid == pid) {
+                return svcConnectToPort(out, service_info->port_h);
+            }
+
+            /* We're mitm'd. */
+            if (R_FAILED(GetMitmServiceHandleImpl(out, service_info, pid))) {
+                /* If the Mitm service is dead, just give a normal session. */
```
misson20000: This should be an std::abort()

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+            /* Validate service name. */
+            R_TRY(ValidateServiceName(service));
+
+            /* Don't try to register something already registered. */
+            if (HasServiceInfo(service)) {
+                return ResultSmAlreadyRegistered;
+            }
+
+            /* Adjust session limit, if compile flags tell us to. */
+#ifdef SM_MINIMUM_SESSION_LIMIT
```
misson20000: One of the things that the Linux kernel does is avoids `#ifdef` in places like these in favor of `if(SM_MINIMUM_SESSION_LIMIT) {`, since the compiler can just remove the code if it's false, but then if someone breaks the code inside, you still get warnings and errors. Maybe we should do that too?

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+            free_service->owner_pid = pid;
+            free_service->max_sessions = max_sessions;
+            free_service->is_light = is_light;
+
+            return ResultSuccess;
+        }
+    }
+
+    /* Initialization. */
+    void InitializeRegistrationLists() {
```
misson20000: Can't this just be done statically?

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+        /* In 8.0.0, Nintendo removed the service apm:p -- however, all homebrew attempts to get */
+        /* a handle to this when calling appletInitialize(). Because hbl has access to all services, */
+        /* This would return true, and homebrew would *wait forever* trying to get a handle to a service */
+        /* that will never register. Thus, in the interest of not breaking every single piece of homebrew */
+        /* we will provide a little first class help. */
+        if (GetRuntimeFirmwareVersion() >= FirmwareVersion_800 && service == ServiceName::Encode("apm:p")) {
+            return ResultSmNotAllowed;
+        }
+
+        /* Check that the process is registered and allowed to register the service. */
```
misson20000:
```suggestion
        /* Check that the process is registered and allowed to get the service. */
```

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+        /* Always clear output. */
+        *out = 0;
+        *out_query = 0;
+
+        /* Create mitm handles. */
+        Handle hnd = 0;
+        Handle qry_hnd = 0;
+        u64 x = 0;
+        R_TRY(svcCreatePort(&hnd, &service_info->mitm_port_h, service_info->max_sessions, service_info->is_light, reinterpret_cast<char *>(&x)));
+        R_TRY_CLEANUP(svcCreateSession(&qry_hnd, &service_info->mitm_query_h, 0, 0), {
```
misson20000: RAII again

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+        if (service_info == nullptr) {
+            return ResultSmNotRegistered;
+        }
+
+        /* Validate that the client pid is the mitm process. */
+        if (service_info->mitm_pid != pid) {
+            return ResultSmNotAllowed;
+        }
+
+        /* Free Mitm session info. */
```
misson20000: Make this into a method on `ServiceInfo`?

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+                }
+            }
+            return nullptr;
+        }
+
+        ProcessInfo *GetFreeProcessInfo() {
+            return GetProcessInfo(InvalidProcessId);
+        }
+
+        void FreeProcessInfo(ProcessInfo *process_info) {
```
misson20000: Maybe move to method on ProcessInfo

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+            out_record->service              = service_info->name;
+            out_record->owner_pid            = service_info->owner_pid;
+            out_record->max_sessions         = service_info->max_sessions;
+            out_record->mitm_pid             = service_info->mitm_pid;
+            out_record->mitm_waiting_ack_pid = service_info->mitm_waiting_ack_pid;
+            out_record->is_light             = service_info->is_light;
+            out_record->mitm_waiting_ack     = service_info->mitm_waiting_ack;
+        }
+
+        void FreeServiceInfo(ServiceInfo *service_info) {
```
misson20000: Same thing here (maybe move to method)

stratosphere/sm/source/sm_user_service.hpp
```diff
@@ -19,46 +19,52 @@
...
-    User_Cmd_UnregisterService = 3,
+namespace sts::sm {
 
-    User_Cmd_AtmosphereInstallMitm = 65000,
-    User_Cmd_AtmosphereUninstallMitm = 65001,
-    User_Cmd_AtmosphereAssociatePidTidForMitm = 65002,
-    User_Cmd_AtmosphereAcknowledgeMitmSession = 65003,
-};
+    /* Command IDs. */
+    enum UserServiceCmd {
```
misson20000: Same thing here- `enum class`?

stratosphere/sm/source/sm_dmnt_service.cpp
```diff
@@ -16,18 +16,22 @@
...
-}
+    Result DmntService::AtmosphereGetRecord(Out<ServiceRecord> record, ServiceName service) {
+        return sm::GetServiceRecord(record.GetPointer(), service);
+    }
 
-void DmntService::AtmosphereGetRecordSize(Out<u64> record_size) {
-    record_size.SetValue(sizeof(SmServiceRecord));
-}
+    void DmntService::AtmosphereListRecords(OutBuffer<ServiceRecord> records, Out<u64> out_count, u64 offset) {
+        R_ASSERT(sm::ListServiceRecords(records.buffer, out_count.GetPointer(), offset, records.num_elements));
```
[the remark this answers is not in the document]
SciresM: ListServiceRecords was a void in the original SM. I made it return a Result here for API parity/to be able to use R_ macros within it, and so I figure why not assert? It should never fail.

stratosphere/sm/source/sm_dmnt_service.hpp
```diff
@@ -19,22 +19,28 @@
...
-        virtual void AtmosphereGetRecordSize(Out<u64> record_size);
-    public:
-        DEFINE_SERVICE_DISPATCH_TABLE {
-            MakeServiceCommandMeta<Dmnt_Cmd_AtmosphereGetRecord, &DmntService::AtmosphereGetRecord>(),
-            MakeServiceCommandMeta<Dmnt_Cmd_AtmosphereListRecords, &DmntService::AtmosphereListRecords>(),
-            MakeServiceCommandMeta<Dmnt_Cmd_AtmosphereGetRecordSize, &DmntService::AtmosphereGetRecordSize>(),
-        };
-};
+    /* Command IDs. */
+    enum DmntServiceCmd {
```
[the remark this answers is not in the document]
SciresM: Yeah, good call. Done locally.

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+            private:
+                const u8 *entry;
+                size_t size;
+            public:
+                AccessControlEntry(const void *e, size_t sz) : entry(reinterpret_cast<const u8 *>(e)), size(sz) {
+                    /* ... */
+                }
+
+                AccessControlEntry GetNextEntry() const {
+                    return AccessControlEntry(this->entry + this->GetSize(), this->size - this->GetSize());
```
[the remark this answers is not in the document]
SciresM: I think I'd prefer to have it as explicitly constructing a new object -- it's nice to have all the functions be const.

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+            /* Acknowledgement members. */
+            bool mitm_waiting_ack;
+            u64 mitm_waiting_ack_pid;
+            Handle mitm_fwd_sess_h;
+        };
+
+        class AccessControlEntry {
+            private:
+                const u8 *entry;
+                size_t size;
```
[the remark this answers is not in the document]
SciresM: Good call.

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+            private:
+                u64 min;
+                u64 max;
+            public:
+                InitialProcessIdLimits() {
+                    if (GetRuntimeFirmwareVersion() >= FirmwareVersion_500) {
+                        /* On 5.0.0+, we can get precise limits from the kernel. */
+                        R_ASSERT(svcGetSystemInfo(&this->min, 2, 0, 0));
+                        R_ASSERT(svcGetSystemInfo(&this->max, 2, 0, 1));
+                    } else {
```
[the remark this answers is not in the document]
SciresM: Okay, I'll change it to >= FirmwareVersion_400. This is wrong in current SM, too, I guess.

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+                        /* On < 5.0.0, we just use hardcoded extents. */
+                        this->min = InitialProcessIdMin;
+                        this->max = InitialProcessIdMax;
+                    }
+
+                    /* Ensure range is sane. */
+                    if (this->min > this->max) {
```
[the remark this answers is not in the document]
SciresM: They do, however PID was allowed to be zero on lower firmwares, and they added that assert only later. We can't assert that min > 0 because min might be == 0 on lower firmwares.

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+                        std::abort();
+                    }
+                }
+
+                bool IsInitialProcess(u64 pid) const {
+                    if (pid == InvalidProcessId) {
+                        std::abort();
```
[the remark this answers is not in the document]
SciresM: Nintendo's equivalent has one here!

I think they got paranoid after smhax happened.

image

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+                static constexpr u64 InitialProcessIdMax = 0x50;
+            private:
+                u64 min;
+                u64 max;
+            public:
+                InitialProcessIdLimits() {
+                    if (GetRuntimeFirmwareVersion() >= FirmwareVersion_500) {
+                        /* On 5.0.0+, we can get precise limits from the kernel. */
+                        R_ASSERT(svcGetSystemInfo(&this->min, 2, 0, 0));
+                        R_ASSERT(svcGetSystemInfo(&this->max, 2, 0, 1));
```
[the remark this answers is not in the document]
SciresM: Yeah. This seems like something to get in libnx, to be honest.

stratosphere/sm/source/sm_service_manager.cpp
```diff
@@ -0,0 +1,723 @@
...
+
+        Result ValidateServiceName(ServiceName service) {
+            /* Service names must be non-empty. */
+            if (service.name[0] == 0) {
+                return ResultSmInvalidServiceName;
+            }
+
+            /* Get name length. */
+            size_t name_len = 1;
+            while (name_len < sizeof(service)) {
```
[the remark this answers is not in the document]
SciresM: yep.

stratosphere/sm/source/sm_service_manager.cpp
SciresM: Yeah, okay.

stratosphere/sm/source/sm_main.cpp
```diff
@@ -108,4 +102,3 @@ int main(int argc, char **argv)
     /* Cleanup. */
 	return 0;
```
fincs: Bad indent

PR #587, opened 2019-06-21
spl: refactor into sts namespace

SciresM:
This refactors the spl sysmodule, placing its functionality into sts::spl::.


stratosphere/spl/source/spl_types.hpp
```diff
@@ -21,141 +21,136 @@
...
+            InvalidArgument       = 2,
+            InProgress            = 3,
+            NoAsyncOperation      = 4,
+            InvalidAsyncOperation = 5,
+            Blacklisted           = 6,
+
+            Max                   = 99,
+        };
+
+        inline ::Result ConvertResult(Result result) {
```
Adubbz: Is `` ::Result`` intentional?

SciresM: Yes, because it returns a Result, not an smc::Result.

PR #606, opened 2019-06-29
pm: complete rewrite

SciresM:
This completely rewrites PM.

Note: Still a bug or two left to track down, right now it fails to process events past boot2 (and so hangs at Nintendo Switch logo).

Still, that should be sorted out soon enough, and making PR now so that code can be reviewed by anyone interested.


stratosphere/pm/source/boot2/boot2_api.cpp
```diff
@@ -0,0 +1,294 @@
...
+namespace sts::boot2 {
+
+    namespace {
+
+        /* Launch lists. */
+
+        /* psc, bus, pcv is the minimal set of required titles to get SD card. */
+        /* bus depends on pcie, and pcv depends on settings. */
+        constexpr ncm::TitleId PreSdCardLaunchPrograms[] = {
+            ncm::TitleId{TitleId_Psc},         /* psc */
```
fincs: Should there be an implicit constructor in `ncm::TitleId` that accepts an u64?

stratosphere/pm/source/boot2/boot2_api.cpp
```diff
@@ -0,0 +1,294 @@
...
+            ncm::TitleId{TitleId_Pcie},        /* pcie */
+            ncm::TitleId{TitleId_Bus},         /* bus */
+            ncm::TitleId{TitleId_Settings},    /* settings */
+            ncm::TitleId{TitleId_Pcv},         /* pcv */
+        };
+        constexpr size_t NumPreSdCardLaunchPrograms = sizeof(PreSdCardLaunchPrograms) / sizeof(PreSdCardLaunchPrograms[0]);
```
fincs: I would suggest adding a template-based countof helper to libstrat.

leoetlino: Any reason for not using [std::size](https://en.cppreference.com/w/cpp/iterator/size) here and for AdditionalLaunchPrograms?

stratosphere/pm/source/boot2/boot2_api.cpp
```diff
@@ -0,0 +1,294 @@
...
+                }
+                svcSleepThread(1'000'000ull);
+            }
+        }
+
+        void LaunchFlaggedProgramsFromSdCard() {
+            /* Allow for user-customizable programs. */
+            DIR *titles_dir = opendir("sdmc:/atmosphere/titles");
+            struct dirent *ent;
+            if (titles_dir != NULL) {
```
leoetlino: (Just a reminder for changing this to nullptr, unless you want to rewrite this further?)

stratosphere/pm/source/impl/pm_resource_manager.cpp
```diff
@@ -0,0 +1,334 @@
...
+        constexpr LimitableResource LimitableResources[] = {
+            LimitableResource_Memory,
+            LimitableResource_Threads,
+            LimitableResource_Events,
+            LimitableResource_TransferMemories,
+            LimitableResource_Sessions,
+        };
+        constexpr size_t LimitableResource_Count = sizeof(LimitableResources) / sizeof(LimitableResources[0]);
+
+        constexpr size_t Megabyte = 0x100000;
```
fincs: Might be worth using [user-defined literals](https://en.cppreference.com/w/cpp/language/user_literal) but that might be slight overkill and I can't think of a suitable namespace in which to place its definition.

stratosphere/pm/source/impl/pm_process_info.hpp
```diff
@@ -0,0 +1,310 @@
...
+
+    };
+
+    Result OnProcessSignaled(std::shared_ptr<ProcessInfo> process_info);
+
+    class ProcessInfoWaiter final : public IWaitable {
+        private:
+            std::shared_ptr<ProcessInfo> process_info;
+        public:
+            ProcessInfoWaiter(std::shared_ptr<ProcessInfo> p) : process_info(p) { /* ... */ }
```
leoetlino: Probably doesn't matter, but:

```suggestion
            ProcessInfoWaiter(std::shared_ptr<ProcessInfo> p) : process_info(std::move(p)) { /* ... */ }
```

stratosphere/pm/source/impl/pm_resource_manager.cpp
```diff
@@ -0,0 +1,334 @@
...
+            g_memory_resource_limits[spl::MemoryArrangement_StandardForSystemDev][ResourceLimitGroup_Applet] -= ExtraSystemMemorySize500;
+        }
+        if (firmware_version >= FirmwareVersion_600) {
+            /* 6.0.0 increased the system event and session limits. */
+            g_resource_limits[ResourceLimitGroup_System][LimitableResource_Events]   += ExtraSystemEventCount600;
+            g_resource_limits[ResourceLimitGroup_System][LimitableResource_Sessions] += ExtraSystemSessionCount600;
+        }
+
+        /* 7.0.0+: Nintendo restricts the number of system threads here, from 0x260 -> 0x60. */
+        /* We will not do this. */
```
fincs: ♫┌( ͡° ͜ʖ ͡°)┘♪ ♫└( ͡° ͜ʖ ͡°)┐♪ 
In line with other places where we decided not to implement a Nintendo restriction (such as anti-dg checks), it might be worth leaving the corresponding code commented out for the sake of reference.

stratosphere/pm/source/impl/pm_resource_manager.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+            g_system_memory_boost_size = boost_size;
+        }
+
+        return ResultSuccess;
+    }
+
+    Result BoostSystemThreadResourceLimit() {
+        /* Starting in 7.0.0, Nintendo reduces the number of system threads from 0x260 to 0x60, */
+        /* Until this command is called to double that amount to 0xC0. */
```
fincs: Same thing here.

stratosphere/pm/source/pm_main.cpp
```diff
@@ -69,30 +69,35 @@ void __libnx_initheap(void) {
...
-    u64 pids[PRIVILEGED_PROCESS_MAX+1];
-    if (R_SUCCEEDED(svcGetProcessList(&num_pids, pids, sizeof(pids)/sizeof(pids[0])))) {
-        for (u32 i = 0; i < num_pids; i++) {
-            const u64 pid = pids[i];
-            if (PRIVILEGED_PROCESS_MIN <= pid && pid <= PRIVILEGED_PROCESS_MAX) {
-                fsprUnregisterProgram(pid);
-                fsprRegisterProgram(pid, pid, FsStorageId_NandSystem,  PRIVILEGED_FAH, sizeof(PRIVILEGED_FAH), PRIVILEGED_FAC, sizeof(PRIVILEGED_FAC));
+namespace {
+
+    static constexpr u32 PrivilegedFileAccessHeader[0x1C / sizeof(u32)]  = {0x00000001, 0x00000000, 0x80000000, 0x0000001C, 0x00000000, 0x0000001C, 0x00000000};
```
fincs: static is redundant with the anonymous namespace.

stratosphere/pm/source/impl/pm_process_manager.cpp
```diff
@@ -0,0 +1,680 @@
...
+        }
+
+        return ResultSuccess;
+    }
+
+    /* Process Management. */
+    Result LaunchTitle(u64 *out_process_id, const ncm::TitleLocation &loc, u32 flags) {
+        /* Ensure we only try to launch one title at a time. */
+        static HosMutex s_lock;
+        std::scoped_lock<HosMutex> lk(s_lock);
```
leoetlino:
```suggestion
        std::scoped_lock lk(s_lock);
```

stratosphere/pm/source/pm_debug_monitor_service.hpp
```diff
@@ -0,0 +1,113 @@
...
+            DEFINE_SERVICE_DISPATCH_TABLE {
+                /* No entries, because DebugMonitorServiceBase is abstract. */
+            };
+    };
+
+    /* This represents modern DebugMonitorService (5.0.0+). */
+    class DebugMonitorService final : public DebugMonitorServiceBase {
+        private:
+            enum class CommandId {
+                GetExceptionProcessIdList        = 0,
```
leoetlino:
```suggestion
                GetExceptionProcessIdList       = 0,
```

stratosphere/pm/source/pm_debug_monitor_service.hpp
```diff
@@ -0,0 +1,113 @@
...
+                MAKE_SERVICE_COMMAND_META(DebugMonitorService, AtmosphereGetCurrentLimitInfo),
+            };
+    };
+
+    /* This represents deprecated DebugMonitorService (1.0.0-4.1.0). */
+    class DebugMonitorServiceDeprecated final : public DebugMonitorServiceBase {
+        private:
+            enum class CommandId {
+                GetModuleIdList                 = 0,
+                GetExceptionProcessIdList        = 1,
```
leoetlino:
```suggestion
                GetExceptionProcessIdList       = 1,
```

stratosphere/pm/source/boot2/boot2_api.cpp
SciresM: It's important that std::is_pod<>::value be true. I think that's incompatible with constructor?

I should make all the TitleId_* definitions be ncm::TitleId though.

stratosphere/pm/source/boot2/boot2_api.cpp
SciresM: Yeah, it'd do well in defines.hpp.

stratosphere/pm/source/impl/pm_resource_manager.cpp
SciresM: Yeah, seems a bit overkill.

stratosphere/pm/source/impl/pm_resource_manager.cpp
SciresM: Hm, good call.

stratosphere/pm/source/impl/pm_resource_manager.cpp
SciresM: Yep.

stratosphere/pm/source/pm_main.cpp
SciresM: Right.

stratosphere/pm/source/boot2/boot2_api.cpp
SciresM: No real reason. I use size_t without std:: all throughout stratosphere. Should I change every instance?

stratosphere/pm/source/boot2/boot2_api.cpp
SciresM: Thank you :)

stratosphere/pm/source/impl/pm_process_info.hpp
SciresM: mm, yeah. Probably doesn't matter but you're right.

stratosphere/pm/source/impl/pm_process_manager.cpp
SciresM: Wait can I really elide that? There's like a million places to fix this if so.

stratosphere/pm/source/pm_debug_monitor_service.hpp
SciresM: Thanks :)

stratosphere/pm/source/pm_debug_monitor_service.hpp
SciresM: Yeah, dunno how I missed these.

stratosphere/pm/source/boot2/boot2_api.cpp
leoetlino: Oh, I meant std::size, not size_t (or std::size_t). std::size is a function that gives you the size of a container or a C-style array.

stratosphere/pm/source/impl/pm_process_manager.cpp
leoetlino: Yep! Class template arguments can now be deduced starting from C++17 :)

stratosphere/pm/source/boot2/boot2_api.cpp
SciresM: ...oh, nice. I had no idea that existed. That's quite useful, thanks! :)

stratosphere/pm/source/impl/pm_process_manager.cpp
SciresM: That's awesome, and good info to have.

stratosphere/pm/source/boot2/boot2_api.cpp
```diff
@@ -0,0 +1,294 @@
...
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        void LaunchTitle(u64 *out_process_id, const ncm::TitleLocation &loc, u32 launch_flags) {
+            u64 process_id = 0;
+
+            /* Don't launch a title twice during boot2. */
```
misson20000: How does this happen? `boot2.flag`? Does Nintendo check this?

stratosphere/pm/source/boot2/boot2_api.cpp
```diff
@@ -0,0 +1,294 @@
...
+            return R_SUCCEEDED(gpioPadGetValue(&button, &val)) && val == GpioValue_Low;
+        }
+
+        bool IsMaintenanceMode() {
+            /* Contact set:sys, retrieve boot!force_maintenance. */
+            DoWithSmSession([&]() {
+                R_ASSERT(setsysInitialize());
+            });
+            {
+                ON_SCOPE_EXIT { setsysExit(); };
```
misson20000: Having separate scopes and separate `DoWithSmSession` blocks for `set:sys` and `gpio` seem to me to be more harmful to readability than they are beneficial, unless the idea is to make it to gpio init errors wouldn't prevent maintenance forced by `set:sys`. The `DoWithSmSession`/`ON_SCOPE_EXIT` pattern doesn't seem to be especially conducive to initializing multiple sessions safely though... here's an idea:
```cpp
template<Result Initializer(), void Finalizer()>
class ServiceHolder {
  public:
    // non-copyable boilerplate
    // move semantics boilerplate
    Result Initialize() {
        if(has_initialized) {
            std::abort(); // sanity
        }
        initialization_result = Initializer();
        if(R_OK(initialization_result)) {
            has_initialized = true;
        }
        return initialization_result; // for convenient R_ASSERT composition
    }
    void Finalize() {
        if(!has_initialized) {
            std::abort(); // invalid state
        }
        Finalizer();
        has_initialized = false;
    }
    void TryFinalize() {
        if(has_initialized) {
            Finalizer();
        }
        has_initialized = false;
    }
    explicit operator bool() const {
        return has_initialized;
    }
    Result GetResult() const {
        return initialization_result;
    }
    ~ServiceHolder() {
        TryFinalize();
    }
  private:
    Result initialization_result;
    bool has_initialized = false;
};
```
````cpp
ServiceHolder<setsysInitialize, setsysExit> setsys_holder;
ServiceHolder<gpioInitialize, gpioExit> gpio_holder;
DoWithSmSession([&]() {
    R_ASSERT(setsys_holder.Initialize());
    R_ASSERT(gpio_holder.Initialize());
});
// check set:sys
// check gpio
````
or,
```cpp
ServiceHolder<setsysInitialize, setsysExit> setsys_holder;
ServiceHolder<gpioInitialize, gpioExit> gpio_holder;
DoWithSmSession([&]() {
    setsys_holder.Initialize();
    gpio_holder.Initialize();
});
R_ASSERT(setsys_holder.GetResult());
// check set:sys, potentially return early and skip second R_ASSERT
R_ASSERT(gpio_holder.GetResult());
// check gpio
```

stratosphere/pm/source/boot2/boot2_api.cpp
```diff
@@ -0,0 +1,294 @@
...
+        void WaitForMitm(const char *service) {
+            const auto name = sts::sm::ServiceName::Encode(service);
+
+            while (true) {
+                bool mitm_installed = false;
+                R_ASSERT(sts::sm::mitm::HasMitm(&mitm_installed, name));
+                if (mitm_installed) {
+                    break;
+                }
+                svcSleepThread(1'000'000ull);
```
misson20000: Can we have `sm` signal an event instead?

stratosphere/pm/source/boot2/boot2_api.cpp
misson20000: `boot2.flag` checks this... https://github.com/Atmosphere-NX/Atmosphere/pull/606/files#diff-0225a52181f3642dc4fc0c7845de571dR234

stratosphere/pm/source/boot2/boot2_api.cpp
```diff
@@ -0,0 +1,294 @@
...
+        void LaunchFlaggedProgramsFromSdCard() {
+            /* Allow for user-customizable programs. */
+            DIR *titles_dir = opendir("sdmc:/atmosphere/titles");
+            struct dirent *ent;
+            if (titles_dir != NULL) {
+                while ((ent = readdir(titles_dir)) != NULL) {
+                    if (strlen(ent->d_name) == 2 * sizeof(u64) && IsHexadecimal(ent->d_name)) {
+                        ncm::TitleId title_id{strtoul(ent->d_name, NULL, 16)};
+                        if (pm::info::HasLaunchedTitle(title_id)) {
+                            return;
```
misson20000: leaks `titles_dir`

stratosphere/pm/source/boot2/boot2_api.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+    }
+
+    /* Boot2 API. */
+    void LaunchBootPrograms() {
+        /* Wait until fs.mitm has installed itself. We want this to happen as early as possible. */
+        WaitForMitm("fsp-srv");
+
+        /* Launch programs required to mount the SD card. */
+        LaunchList(PreSdCardLaunchPrograms, NumPreSdCardLaunchPrograms);
```
misson20000: Agree with `std::size` sentiment above.
```suggestion
        LaunchList(PreSdCardLaunchPrograms, std::size(PreSdCardLaunchPrograms));
```

stratosphere/pm/source/impl/pm_process_info.hpp
```diff
@@ -0,0 +1,310 @@
...
+            }
+
+            void SetExceptionOccurred() {
+                this->SetFlag(Flag_ExceptionOccurred);
+                this->SetFlag(Flag_ExceptionWaitingAttach);
+            }
+
+            void ClearExceptionOccurred() {
+                this->ClearFlag(Flag_ExceptionOccurred);
+            }
```
misson20000: lots of boilerplate with these flag setters/clearers... macro maybe?

stratosphere/pm/source/impl/pm_process_info.hpp
```diff
@@ -0,0 +1,310 @@
...
+
+            void Add(std::shared_ptr<ProcessInfo> process_info) {
+                this->processes.push_back(process_info);
+            }
+
+            void Remove(u64 process_id) {
+                for (size_t i = 0; i < this->processes.size(); i++) {
+                    if (this->processes[i]->GetProcessId() == process_id) {
+                        this->processes.erase(this->processes.begin() + i);
```
misson20000: use an iterator to avoid `this->processes.begin() + i`?

stratosphere/pm/source/impl/pm_process_info.hpp
```diff
@@ -0,0 +1,310 @@
...
+                        break;
+                    }
+                }
+            }
+
+            std::shared_ptr<ProcessInfo> Find(u64 process_id) {
+                for (size_t i = 0; i < this->processes.size(); i++) {
```
misson20000: These should probably use iterators too

stratosphere/pm/source/impl/pm_process_manager.cpp
```diff
@@ -0,0 +1,680 @@
...
+        /* Hook globals. */
+        std::atomic<ncm::TitleId> g_title_id_hook;
+        std::atomic<bool> g_application_hook;
+
+        /* Helpers. */
+        void ProcessTrackingMain(void *arg) {
+            /* This is the main loop of the process tracking thread. */
+
+            /* Create waitable manager. */
+            static auto s_process_waiter = WaitableManager(1);
```
misson20000: Why are you using auto here?
```
static WaitableManager s_process_waiter(1);
```

stratosphere/pm/source/impl/pm_process_manager.cpp
```diff
@@ -0,0 +1,680 @@
...
+            g_process_waitable_manager = &s_process_waiter;
```
misson20000: Why is this a global pointer instead of just making the `WaitableManager` itself global and have things access it directly? Are you trying to defer the initialization of `WaitableManager` until `ProcessTrackingMain` is called?

stratosphere/pm/source/impl/pm_process_manager.cpp
```diff
@@ -0,0 +1,680 @@
...
+
+        Result LaunchProcess(const LaunchProcessArgs *args) {
+            /* Get Program Info. */
+            ldr::ProgramInfo program_info;
+            R_TRY(ldr::pm::GetProgramInfo(&program_info, args->location));
+            const bool is_application = (program_info.flags & ldr::ProgramInfoFlag_ApplicationTypeMask) == ldr::ProgramInfoFlag_Application;
+            const bool allow_debug    = (program_info.flags & ldr::ProgramInfoFlag_AllowDebug) || GetRuntimeFirmwareVersion() < FirmwareVersion_200;
+
+            /* Ensure we only try to run one application. */
+            if (is_application) {
```
misson20000: Uh... is this correct? Shouldn't it be `if (is_application && /* already has application */) {` ?

stratosphere/pm/source/impl/pm_process_manager.cpp
```diff
@@ -0,0 +1,680 @@
...
+
+    }
+
+    /* Initialization. */
+    Result InitializeProcessManager() {
+        /* Create events. */
+        g_process_event = CreateWriteOnlySystemEvent();
+        g_hook_to_create_process_event = CreateWriteOnlySystemEvent();
+        g_hook_to_create_application_process_event = CreateWriteOnlySystemEvent();
+        g_boot_finished_event = CreateWriteOnlySystemEvent();
```
misson20000: Transferring ownership by moving pointers around isn't very C++, but that's more of a general stratosphere concern than a PM concern.

stratosphere/pm/source/impl/pm_process_manager.cpp
```diff
@@ -0,0 +1,680 @@
...
+            return ResultPmProcessNotFound;
+        }
+
+        return svcTerminateProcess(process_info->GetHandle());
+    }
+
+    Result TerminateTitle(ncm::TitleId title_id) {
+        ProcessListAccessor list(g_process_list);
+
+        auto process_info = list->Find(title_id);
```
misson20000: beware adding an implicit constructor to `ncm::TitleId` and messing up this overload

stratosphere/pm/source/pm_main.cpp
```diff
@@ -69,30 +69,35 @@ void __libnx_initheap(void) {
...
+    static constexpr u32 PrivilegedFileAccessControl[0x2C / sizeof(u32)] = {0x00000001, 0x00000000, 0x80000000, 0x00000000, 0x00000000, 0xFFFFFFFF, 0xFFFFFFFF, 0x00000000, 0x00000000, 0xFFFFFFFF, 0xFFFFFFFF};
+    static constexpr size_t ProcessCountMax = 0x40;
+
+    /* This works around a bug fixed by FS in 4.0.0. */
+    /* Not doing so will cause KIPs with higher process IDs than 7 to be unable to use filesystem services. */
```
misson20000: Appreciate this comment :+1:

stratosphere/pm/source/pm_main.cpp
```diff
@@ -69,30 +69,35 @@ void __libnx_initheap(void) {
...
+        /* Get privileged process range. */
+        u64 min_priv_process_id = 0, max_priv_process_id = 0;
+        sts::cfg::GetInitialProcessRange(&min_priv_process_id, &max_priv_process_id);
+
+        /* Get list of processes, register all privileged ones. */
+        u32 num_pids;
+        u64 pids[ProcessCountMax];
+        R_ASSERT(svcGetProcessList(&num_pids, pids, ProcessCountMax));
+        for (size_t i = 0; i < num_pids; i++) {
+            if (min_priv_process_id <= pids[i] && pids[i] <= max_priv_process_id) {
```
misson20000: If you do this before launch any processes, can you just assume they're all KIPs and therefore privileged?

stratosphere/pm/source/boot2/boot2_api.cpp
SciresM: Oh, so it does. Nice, I'll remove the check then.

stratosphere/pm/source/boot2/boot2_api.cpp
SciresM: I love this idea. However, let's do it like AutoCloseMap -- constructor sets this->result = Initialize(). Destructor closes if Result was success -- don't require member calls to init, so that object scope is directly tied to service state.

stratosphere/pm/source/boot2/boot2_api.cpp
SciresM: Events are kind of a no, since we don't want to make one for every service. However, I could simply add a WaitMitm command or something like that? SM can defer result return until ready. Thoughts?

stratosphere/pm/source/boot2/boot2_api.cpp
SciresM: Oh, good catch.

stratosphere/pm/source/boot2/boot2_api.cpp
SciresM: Yep.

stratosphere/pm/source/impl/pm_process_info.hpp
SciresM: Yeah, I thought about it. I'll see how it looks?

stratosphere/pm/source/impl/pm_process_info.hpp
SciresM: ah, yeah, good call.

stratosphere/pm/source/impl/pm_process_info.hpp
SciresM: Any reason why?

stratosphere/pm/source/impl/pm_process_manager.cpp
SciresM: WaitableManager isn't an instantiable class because templating. It's actually WaitableManager<DefaultOptions> or something. I use auto for that reason, since I prefer it to SessionManagerBase.

stratosphere/pm/source/impl/pm_process_manager.cpp
SciresM: Because the WaitableManager manager must not be initialized until after __appInit is called.

stratosphere/pm/source/impl/pm_process_manager.cpp
SciresM: Haha, yeah, good catch.

stratosphere/pm/source/impl/pm_process_manager.cpp
SciresM: Yeah, problem as always is that I can't statically initialize these :/

It's kind of a design issue. Is there any way to make __appInit happen before static initializers are called?

SciresM: Pinging @fincs

stratosphere/pm/source/impl/pm_process_manager.cpp
SciresM: ncm::TitleId must never have a constructor of any kind, because it must be pod.

stratosphere/pm/source/pm_main.cpp
SciresM: Yep, glad it's useful.

stratosphere/pm/source/pm_main.cpp
SciresM: You can, I guess, yeah. I kind of prefer the check but it is admittedly superfluous.

stratosphere/pm/source/impl/pm_process_manager.cpp
fincs: [__appInit already happens before static initializers](https://github.com/switchbrew/libnx/blob/master/nx/source/runtime/init.c#L177-L182)

SciresM: Okay, cool. I'll statically initialize these then.

stratosphere/pm/source/boot2/boot2_api.cpp
misson20000: Not sure I made clear what I meant in Discord and couldn't reply here on mobile, but anything to avoid the spin loop. I meant have a single event that signals whenever any mitm service comes up and keep the loop, just replace the `svcSleepThread` with an event wait. `WaitMitm` works too.

stratosphere/pm/source/impl/pm_process_info.hpp
misson20000: Consistency with Remove (if you change that to use iterators), and iterators are generally considered better style when working with STL containers (although `<algorithm>`s like `std::find` are best style, but the whole predicate thing is kinda bothersome).

stratosphere/pm/source/impl/pm_process_manager.cpp
misson20000: Any changes to this after @fincs confirmation that `_appInit` happens before static initializers?

SciresM: Yeah, this will be addressed, too.

PR #616, opened 2019-07-01
Fix for 4.X units (new strato rework broke a method in spl)

ckurtz22:
The stratosphere rework looks super clean but it broke one case where there was an alternate function in spl that is only for 4.X. The CommandId is the same but the function to call on 4.X is a slightly different name, which likely caused it to be missed when comparing stuff, and it is only used on 4.X so hard to notice the issue (the issue being it won't boot).

Probably not the cleanest fix to this, but this solution does fix the issue.


SciresM:
Ha. Yeah, that would do it. Thanks!

Could you fix the alignment of = in the enum?

ckurtz22:
Done

SciresM:
Excellent, thank you.

kronenbierski:
Can you please tell me how to fix it? I am having the same problem.

Raugo:
Download the latest release of atmosphere, was updated 2 days ago to include it

ValliantStorme:
@kronenbierski This isn't a support forum, but the solution is redownloading the latest Atmosphere release

kronenbierski:
Thank you both for your answers. I did download the latest release, but
after the atmosphere logo i get a black screen. Will continue to search for
a solution. Since I am doing this on another pc i will install driver with
zadig next, causecthat is the only thing i haven't tried.

PR #630, opened 2019-07-16
defaults: disable screenshot verification by capsrv

ThatNerdyPikachu:


common/defaults/system_settings.ini
```diff
@@ -8,6 +8,9 @@ usb30_force_enabled = u8!0x0
 ; (note: this is normally not necessary, and ips patches can be used.)
 [ro]
 ease_nro_restriction = u8!0x0
+; Disable screenshot verification
+[capsrv]
+enable_album_screenshot_filedata_verification = u8!0x1
```
misson20000: Are you sure this is correct? 1 sure looks a lot like "yes please do verify screenshots" (and is also the default...)

ThatNerdyPikachu:
```
capsrv is responsible for validating the MACs for screenshots stored on the SD card, if the setting capsrv!enable_album_screenshot_filedata_verification is non-zero. 
```

- [switchbrew](https://switchbrew.org/wiki/Capture_services#Notes)

misson20000: So if it's non-zero, it verifies MACs, so you need to set it zero.

ThatNerdyPikachu: oh, i can read.

SciresM:
Why is this a desirable default, over having a third party program that correctly signs screenshots? The latter seems preferable to me.

SciresM:
To add on to that -- Atmosphere doesn't typically remove or patch validity checks -- we still require that NSOs have valid hashes, and that NROs have a hash present in the NRR they're associated to. 

Given that this MAC is a hash (symmetrically producable by the user) rather than e.g. an asymmetric signature, I'm not sure I see the benefit?

ThatNerdyPikachu:
> Why is this a desirable default, over having a third party program that correctly signs screenshots? The latter seems preferable to me.

To make it easier on the user?

SciresM:
But why would that be a desirable default? The typical user isn't trying to view custom photos/videos in his/her album applet (is there a homebrew to display photos/videos?)

I'm not a fan of disabling validation in general, and I'm not sure I understand why we'd want to make the system not validate the correctness of a file when it's possible for us to forge correct files.

PR #634, opened 2019-07-21
Enable ccache

fennectech:
This pull enables ccache for atmosphere compilation on machines with ccache in place making it faster to compile.  This does nothing on machines without ccache.   I see no reason not to have it.  This no effect on output code.


SciresM:
Ams already only rebuilds files on change?

fincs:
Please see https://github.com/switchbrew/libnx/pull/311#issuecomment-513542300

PR #643, opened 2019-08-02
New service for USB drive access: fsp-usb

XorTroll:
The service worked fine as a standalone process, but is having a few issues as part of ams-mitm. Pull request opened for code reviews, since this issues need to be worked out.


stratosphere/ams_mitm/source/fsp_usb/fspusb_main.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <malloc.h>
+
+#include <switch.h>
+#include <atmosphere.h>
+#include <stratosphere.hpp>
+
+#include "fspusb_main.hpp"
+#include "fspusb_service.hpp"
+
+// fsp-srv's ones, but without domains (need this specific buf size as we're using fsp-like IFileSystems)
```
SciresM: Actual in-depth review later, but this needs to support domains.

(Also, use /* comment */ over // comment)

rhyskoedijk:
Apologies in advance, this is very off-topic but I don't know how else to contact you directly.

@XorTroll, this looks really great. do you have any plan to expose this kind of mass-storage as a separate library? I'm doing a port of NTFS-3G over in one of my repos. It currently compiles, but without underlying block access to the device, it doesn't currently do anything. I was hoping to maybe use your code to achieve this (unless you are already doing it). It only need a very basic interface, something like: 
https://github.com/rhyskoedijk/libstorage-nx/blob/master/include/storage.h

fincs:
I'm strongly of the opinion that if all, Atmosphère should expose a generic pluggable IFileSystem interface registry instead of hardcoding support for a particular storage medium + filesystem format.

rhyskoedijk:
@XorTroll Just curious if https://github.com/switchbrew/libnx/issues/337 is in any way part of the issues you ran in to? Might not, but I just saw that issue reported today and in hindsight this explains one of the crashes I kept running in to doing my own USB mass storage testing when closing endpoints.

I'm really looking forward to seeing this service get implemented at some point.  😃

XorTroll:
I started rewriting fsp-usb like 2 weeks ago, and it already works way better than the original code. I've also tested it within ams_mitm, and works fine too.
Might close this PR and reopen it with the new branch containing the updated code.

XorTroll:
https://github.com/XorTroll/Atmosphere/tree/fspusb

XorTroll:
Closing this PR since fsp-usb development is active on a different branch, and there's still a lot to do

PR #669, opened 2019-09-19
Changed reboot logic

Mika-Westphal:
I have replaced the reboot logic with the new bpcRebootSystem function, because it doesn't need as much code as the other method and it's way easier to understand.


fincs:
This will not work for the intended purpose, i.e. forcing a reread of `reboot_payload.bin`. ams_mitm's BpcRebootManager object only reads the payload during sysmodule initialization.

Mika-Westphal:
oh ok sorry ^^

PR #697, opened 2019-11-13
Allow system, application and applet memory to be set manually

TheGreatRambler:
In `system_settings.ini`, you can now set memory manually through these settings:

`atmosphere:manual_memory_control`: Whether to read the memory values manually or generate them based on various hardware details
`atmosphere:system_memory`: System memory in Mb
`atmosphere:application_memory`: Application memory in Mb
`atmosphere:applet_memory`: Applet memory in Mb

These settings are marked as advanced and dangerous.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,3 @@
-# WHY DOES THIS FORK EXIST
-All this fork does is allows the size of the System memory pool to be changed
-
-## *START OF ORIGINAL README*
-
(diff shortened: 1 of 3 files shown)


SciresM:
This code does not work and cannot work. PM does not have the ability to access set:sys, because resource limits must be set before set:sys is launched.

PR #744, opened 2019-12-09
Vapourware

SciresM:
This moves libstratosphere to /libraries, and sets up for common atmosphere libraries in the future.

It also generally greatly improves stratosphere's buildsystem.


SciresM:
This will be merged after the 0.10.0 release stops being a pre-release.

SciresM:
Turns out we're merging this now! This is because 0.10.1 is pretty likely gonna happen in the next day or so (as more bugs were fixed than expected).

PR #760, opened 2019-12-13
loader: fix ECS after new-ipc rewrite

misson20000:
The call to serviceCreate(...) tries to query pointer buffer size, but
since we haven't had a chance to return the server side of the session
yet, this deadlocks. Instead, we defer creating the session and
mounting the filesystem until the first time the ECS object is used.

### DIFF
--- a/stratosphere/loader/source/ldr_ecs.cpp
+++ b/stratosphere/loader/source/ldr_ecs.cpp
@@ -28,15 +28,41 @@ namespace ams::ldr::ecs {
             NON_COPYABLE(ExternalContentSource);
             NON_MOVEABLE(ExternalContentSource);
             private:
+                bool has_mounted = false;
                 char device_name[DeviceNameSizeMax];
+                os::ManagedHandle client;
+
+                Result Mount() {
+                    /* Create service. */
+                    Service srv;
+                    serviceCreate(&srv, client.Move());
+                    FsFileSystem fs = { srv };
+                    auto fs_guard = SCOPE_GUARD { fsFsClose(&fs); };
+
+                    /* Try to mount. */
+                    R_UNLESS(fsdevMountDevice(device_name, fs) >= 0, fs::ResultMountNameAlreadyExists());
+                    fs_guard.Cancel();
+
+                    this->has_mounted = true;
+                    return ResultSuccess();
+                }
             public:
-                ExternalContentSource(const char *dn) {
+                ExternalContentSource(const char *dn, os::ManagedHandle client) : client(std::move(client)) {
                     std::strncpy(this->device_name, dn, sizeof(this->device_name));
                     this->device_name[sizeof(this->device_name) - 1] = '\0';
                 }
 
+                Result EnsureMounted() {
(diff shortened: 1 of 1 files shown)


stratosphere/loader/source/ldr_ecs.cpp
```diff
@@ -28,15 +28,40 @@ namespace ams::ldr::ecs {
...
+                }
             public:
-                ExternalContentSource(const char *dn) {
+                ExternalContentSource(const char *dn, os::ManagedHandle client) : client(std::move(client)) {
                     std::strncpy(this->device_name, dn, sizeof(this->device_name));
                     this->device_name[sizeof(this->device_name) - 1] = '\0';
                 }
 
+                void EnsureMounted() {
+                    if(!has_mounted) {
```
SciresM: this->has_mounted

stratosphere/loader/source/ldr_ecs.cpp
```diff
@@ -28,15 +28,40 @@ namespace ams::ldr::ecs {
...
+                        Mount();
+                    }
+                }
+
                 ~ExternalContentSource() {
-                    fsdevUnmountDevice(this->device_name);
+                    if(has_mounted) {
```
SciresM: this->has_mounted, space between if and (

stratosphere/loader/source/ldr_ecs.cpp
SciresM: also, space between if and (

stratosphere/loader/source/ldr_ecs.cpp
```diff
@@ -28,15 +28,40 @@ namespace ams::ldr::ecs {
...
+                    Service srv;
+                    serviceCreate(&srv, client.Move());
+                    FsFileSystem fs = { srv };
+                    auto fs_guard = SCOPE_GUARD { fsFsClose(&fs); };
+
+                    /* Try to mount. */
+                    R_UNLESS(fsdevMountDevice(device_name, fs) >= 0, fs::ResultMountNameAlreadyExists());
+                    fs_guard.Cancel();
+
+                    has_mounted = true;
```
SciresM: this->has_mounted

stratosphere/loader/source/ldr_ecs.cpp
```diff
@@ -28,15 +28,40 @@ namespace ams::ldr::ecs {
...
             public:
-                ExternalContentSource(const char *dn) {
+                ExternalContentSource(const char *dn, os::ManagedHandle client) : client(std::move(client)) {
                     std::strncpy(this->device_name, dn, sizeof(this->device_name));
                     this->device_name[sizeof(this->device_name) - 1] = '\0';
                 }
 
+                void EnsureMounted() {
+                    if(!has_mounted) {
+                        Mount();
```
SciresM: What happens in the case where Mount() fails?

misson20000: We now silently remove the ECS from the map and pretend it never existed.

PR #761, opened 2019-12-13
pm: fix busy loop on <5.0.0 when processes exit

### LINKED ISSUE #748: hbmenu freeze on 0.10
im running atmosphere 0.10.0 on firmware 4.0.1 and when in hbmenu and i push the home button my switch will lockup and i have to hold the power button for 12 seconds to turn it off 
there is no crash report

misson20000:
Fixes #748. This is a bit of a "good enough" fix. It works, and proves what the problem is, but you may prefer to reorganize the management of whether the waitable holder is linked or not so that it does not require a flag.


SciresM:
Yeah I wouldn't merge this because it makes lifetimes ambiguous.

Good to know that's the problem.

SciresM:
Closed by commit to master without lifetime ambiguity. Thanks :)

PR #770, opened 2019-12-17
Mesosphere: Implement KernelLdr (8.0.0+)

SciresM:
This implements kernel_ldr, a binary added in 8.0.0 which applies relocations to the kernel and maps it at a random base address. It also sets up some hardware registers.

This PR also skeletons a buildsystem for the new libmesosphere, which will hold all common code between kernel_ldr and the kernel we will be reimplementing in the coming year.


libraries/libmesosphere/source/board/nintendo/switch/kern_secure_monitor.hpp
```diff
@@ -0,0 +1,72 @@
...
+namespace ams::kern::smc {
+
+    /* Types. */
+    enum MemoryMode {
+        MemoryMode_4GB = 0,
+        MemoryMode_6GB = 1,
+        MemoryMode_8GB = 2,
+    };
+
+    enum class ConfigItem : u32 {
```
hexkyz: Maybe double check the names here and decide a final scheme (e.g.: NewHardwareType is RegulatorType).

SciresM: Good call :)

PR #875, opened 2020-04-12
erpt: reimplement the sysmodule

SciresM:
This is a complete, functional implementation of the `erpt` sysmodule, tested and working on latest firmware.

Our erpt saves ~0.25 MB (0x48000) memory-wise over Nintendo's implementation. 

Atmosphere extensions are minimal, but present:
* Existing reports are managed exactly as they would be managed by official erpt sysmodule.
* New reports are instead saved to the SD card and not committed to the journal of reports the sysmodule manages.
  * This means that all future error reports will generate no telemetry to be sent to Nintendo.

Only remaining task before merge is more testing, and deciding if we should do anything special about attachments -- right now they're treated identically to official, with the caveat that all attachments for a new report are deleted instead of being stored in the journal. Maybe we want to save new attachments to SD card?


ShadowOne333:
Interesting.
So this would be a way to mitigate all telemetry on the Switch and an implementation of the PRODINFO commit #546 ?
Also, would there be a way to delete/redirect currently existing reports on the console?

shchmue:
this does nothing to prevent games or sysmodules from creating prepo or srepo reports and nothing to keep those modules from sending them, it’s just about error reports.

Issue #874, opened 2020-04-11
[dmntcht] Expose svcContinueDebugEvent and svcBreakProcess functions


olliz0r:
#### Overview:
- Extend dmntcht to allow other sysmodules/homebrew to call svcBreakProcess and svcContinueDebugEvent on the debugged process</br>

#### Smaller Details: 
- N/A</br>

#### Nature of Request:  
- Addition
  - Extending the dmntcht service by two functions, exposing the svcBreakProcess and svcContinueDebugEvent functions on the debughandle.

### What component do you feel this would best fit within?
- [Stratosphère](https://github.com/Atmosphere-NX/Atmosphere#components)</br>
    - dmnt</br>

### Why would this feature be useful?
This would allow homebrew applications and other sysmodules to freeze and unfreeze the process before writing to/reading from it to guarantee, that the application does not modify the data before the operation has completed. 

Example: Using dmntcht right now it is not possible to inject a full box of pokemon in pokemon sword/shield using a sysmodule while the game is running, without the game noticing half written slots and "fixing" them, resulting in broken data (bad eggs). Going the "original" route of calling svcDebugActiveProcess, svcWriteDebugProcessMemory and CloseHandle works fine, however, because the process stays frozen during the writing.</br>

SciresM:
https://github.com/Atmosphere-NX/Atmosphere/pull/899

SciresM:
Please verify that this works/meets your needs, @olliz0r.

SciresM:
Also pinging @AnalogMan151 in case he has any interest in testing.

SciresM:
Verified to work by analogman.

PR #899, opened 2020-04-17
Dmnt: Add break/continue commands, add static register api.

SciresM:
This addresses both the request in #874 as well as a request @WerWolv brought to me in private.

It will be merged following confirmation by the involved parties.


macserv:
You may also want to update `docs/modules/dmnt.md` accordingly.

PR #910, opened 2020-04-19
Add ini option to disable cal0 write protection

NickHu:
This pull request makes write-protection on cal0 configurable by ini. In particular, this is desirable so that the user can blank their PRODINFO on their emunand, which practically severs the connection to Nintendo's servers.

I've tried to put up sufficiently many warnings to dissuade the average user from toggling it without being aware of the repercussions.

### DIFF
--- a/config_templates/system_settings.ini
+++ b/config_templates/system_settings.ini
@@ -26,6 +26,9 @@
 ; Enable reading the CAL0 partition for HBL.
 ; This is probably undesirable for normal usage.
 ; enable_hbl_cal_read = u8!0x0
+; Disable CAL0 write-protection
+; This is almost certainly undesirable and dangerous for normal usage
+; disable_cal_write = u8!0x0 WARRANTY VOID IF CHANGED - YOU HAVE BEEN WARNED
 ; Controls whether fs.mitm should redirect save files
 ; to directories on the sd card.
 ; 0 = Do not redirect, 1 = Redirect.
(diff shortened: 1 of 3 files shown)


SciresM:
I am not willing to merge this. There's really no good reason to write garbage to one's PRODINFO.

I am, however, willing to clean up and modernize #546. This would add an additional system setting that would cause the settings sysmodule to read a blanked PRODINFO without actually modifying the data in NAND. This is actually reasonable to do now that atmosphere updates no longer overwrite user configuration files. 

The other reasoning for not merging that PR before was that there wad no research done demonstrating info from PRODINFO wasn't cached elsewhere (in fact, it is cached elsewhere in several places). However, I am willing to add the blanking as a system setting with a huge warning surrounding it -- it's better than people willingly corrupting their NANDs.

Does this satisfy your use case? If so, I'll get around to that this week (it would be included in Atmosphere 0.11.2).

SciresM:
(A sane implementation of this probably has separate settings for emummc/sysmmc).

NickHu:
@SciresM I agree with you that #546 is a much better approach. The only reason I tried to upstream this pull request is it seems other people are doing this anyway, which has potentially severe reprocussions for the average user who doesn't know or care about why their PRODINFO is important. Having mainline support for it could've helped protect them a bit more, but #546 supercedes this.

SciresM:
Implemented via #913

PR #912, opened 2020-04-19
jpegdec reimplementation

HookedBehemoth:


SciresM:
Nice. @HookedBehemoth Have you tested this as functional on various system versions? In particular wondering about the cases before jpegdec sysmodule existed. Guess code needs added to boot2 to not launch this on < 4.0.0 or whatever.

Also, in matching with the style of recent sysmodule implementations *most* of this code should be inside libstratosphere.

I will probably refactor it heavily to move it in there following merge (will also add e.g. static service object init the way N does it).

SciresM:
Also, please note (per my cleanup commit):

Atmosphere code should *never* do e.g. Result rc = (...), except in very exceptional cases.

R_TRY(), R_UNLESS(), scope guards are your friend.

Also, *never* skip braces around if statements.

PR #916, opened 2020-04-23
Add switch-jpeg-turbo to the list of dependencies.

jhNsXO:
add dependency needed since 90d754f920e099d581d250145eeb00e83303dac0


HookedBehemoth:
Is freetype still required?
fatal uses stb truetype by now.

jhNsXO:
Oh, you're right. I don't see it referenced anywhere else too. Currently testing it without, but compiling takes like 2h on my machine.

PR #930, opened 2020-04-27
Update FatFs to R0.14 patch 1

Felix10R:


hexkyz:
These changes don't seem to have been tested. All of the sdmmc hooks have been replaced by what seems to have been a blind copy paste from the original FatFs repository?

Felix10R:
You're right and I apologize for that. I haven't thought about any changes being made to that. I've had the intention of opening up a pull request instead of an issue to shorten the time of updating it. I've recently had multiple crashes while opening hbmenu, which were simply fixed by overwriting all of the Atmosphere files. My only thoughts for the occurrence of the crashes were possible connections with FatFs. So i tried updating it.

shchmue:
your issues are more likely due to SD problems; the card is getting old, it's fake, it's using exFAT which is prone to corruption, or the archive bits are getting messed up by mounting it with write access in an archive-bit-aggressive OS like Mac OS

Felix10R:
It's a 1 year old SanDisk Ultra 400 GB formatted in exFAT and written using Windows.

masagrator:
Then don't use exFat. Nintendo driver sucks for anything outside of official usage.

Felix10R:
All right. Thanks for the explanations.

PR #935, opened 2020-04-28
Added some documentation for the FSS0 format.

HarukoNX:


SciresM:
1) this documentation isn't even right, the FSS0 content header isn't at a fixed offset
2) The offset for latest build is 0xA0, not 0x400
3) FSS0 is kind of an unofficial/changing format for hekate, not really sure that I want official documentation for it as that implies coherent support to non-hekate clients...

SciresM:
fwiw for your own info, fss0 content header is pointed to by the u32 at fusee_secondary crt0 + 4

PR #958, opened 2020-05-17
ncm: implement firmware downgrading

Adubbz:
Implements the foundations necessary for firmware downgrading.

### DIFF
--- a/libraries/libstratosphere/source/ncm/ncm_package_system_downgrade_task.cpp
+++ b/libraries/libstratosphere/source/ncm/ncm_package_system_downgrade_task.cpp
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2019-2020 Adubbz, Atmosphère-NX
+ *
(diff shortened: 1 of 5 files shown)


libraries/libstratosphere/source/ncm/ncm_package_system_downgrade_task.cpp
```diff
@@ -0,0 +1,76 @@
...
+            /* Obtain the content meta. */
+            InstallContentMeta content_meta;
+            R_TRY(data.Get(std::addressof(content_meta), i));
+
+            /* Create a reader. */
+            const auto reader = content_meta.GetReader();
+            const auto key = reader.GetKey();
+
+            /* Obtain a list of suitable storage ids. */
+            auto storage_list = GetStorageList(this->GetInstallStorage());
```
VolcaEM, line 40: This can be made const (I think)

PR #1037, opened 2020-06-15
added lz4.block

GametasticYT:
without adding this, it gives error "AttributeError: module 'lz4' has no attribute 'block'" when compiling


SciresM:
This isn't correct, you should install the correct version of the lz4 python module (same one as needed for nxo64.py ida loader).

ndeadly:
@SciresM what _is_ the correct version for the ida loader? requirements.txt just says `lz4>=0.10.0`. I'm bumping into the same error with both ida and my system python reporting version 1.1.0

PR #1039, opened 2020-06-16
New Lz4 Api

Kronos2308:
print ( )
work on python 3.8 for me


SciresM:
Won't accept the additional import because it's not necessary using correct lz4 install version. Make another PR with the () fixes.

Kronos2308:
Ok

PR #1048, opened 2020-06-17
Several improvements to split_program.py

edgeyboy:
This adds python2 support by allowing old lz4 syntax (latest lz4 library available for python2 still seems to require the old syntax), and adds python3 support by fixing print syntax.


SciresM:
"This adds python2 support" what? I use python2 with the current script.

SciresM:
Closing due to the fact that it still works fine with py2 here...

PR #1078, opened 2020-07-09
ncm: Mark two variables as const

VolcaEM:
ncm: Mark two variables as const


SciresM:
Thanks, I guess?

For future reference, just DM me something trivial like this on discord.

PR #1083, opened 2020-07-11
daybreak: Added support for factory resetting

Adubbz:
As per title.


SciresM:
Rebase onto master instead of doing triple merge commits, please.

PR #1089, opened 2020-07-16
Add stack overflow detection to fatal error code

misson20000:
Saves some two minutes of looking at a fatal error report and a disassembly to figure out that something daborted because the stack overflowed on a sequence like this and that you might need to increase stack size:
```
sub sp, sp, 0x4f0
stp x29, x30, [sp]
```


SciresM:
& -> std::addressof? Other than that, sure.

misson20000:
Sure thing. Also ported over one more instance of `svcQueryMemory` that I'd missed earlier.

PR #1098, opened 2020-07-27
Fix building with devkitarm-rules 1.1.0

hthh:
The Atmosphere build was broken by https://github.com/devkitPro/devkitarm-rules/commit/5325c8bd6d43dcb4029c62e1d482e51002c6c956

This fix is backwards compatible with earlier devkitarm-rules, as the `_size` names were already defined in the headers generated by the Makefile.


SciresM:
Thanks :)

PR #1125, opened 2020-08-13
Kernel: Improve genericity of code.

SciresM:
This makes porting the kernel to other boards easier.

### DIFF
--- a/libraries/config/arch/arm64/cpu/cortex_a57/cpu.mk
+++ b/libraries/config/arch/arm64/cpu/cortex_a57/cpu.mk
@@ -1,5 +1,5 @@
 export ATMOSPHERE_DEFINES  += -DATMOSPHERE_CPU_ARM_CORTEX_A57
-export ATMOSPHERE_SETTINGS += -mtune=cortex-a57
+export ATMOSPHERE_SETTINGS += -march=armv8-a+crc+crypto -mtune=cortex-a57
 export ATMOSPHERE_CFLAGS   +=
 export ATMOSPHERE_CXXFLAGS +=
 export ATMOSPHERE_ASFLAGS  +=
--- a/libraries/libmesosphere/include/mesosphere/kern_common.hpp
+++ b/libraries/libmesosphere/include/mesosphere/kern_common.hpp
@@ -22,6 +22,12 @@ namespace ams::kern {
 
     constexpr size_t PageSize = 4_KB;
 
+#ifdef ATMOSPHERE_BOARD_NINTENDO_NX
     ams::TargetFirmware GetTargetFirmware();
+#else
+    consteval ALWAYS_INLINE ams::TargetFirmware GetTargetFirmware() {
+        return ams::TargetFirmware_Current;
+    }
+#endif
 
 }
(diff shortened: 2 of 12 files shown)


libraries/config/arch/arm64/cpu/cortex_a57/cpu.mk
```diff
@@ -1,5 +1,5 @@
 export ATMOSPHERE_DEFINES  += -DATMOSPHERE_CPU_ARM_CORTEX_A57
-export ATMOSPHERE_SETTINGS += -mtune=cortex-a57
+export ATMOSPHERE_SETTINGS += -march=armv8-a+crc+crypto -mtune=cortex-a57
```
SciresM, line 2: Should this be a board thing? I guess this is the right place to put it...

libraries/libmesosphere/include/mesosphere/kern_common.hpp
```diff
@@ -22,6 +22,12 @@ namespace ams::kern {
 
     constexpr size_t PageSize = 4_KB;
 
+#ifdef ATMOSPHERE_BOARD_NINTENDO_NX
     ams::TargetFirmware GetTargetFirmware();
+#else
+    constexpr ALWAYS_INLINE ams::TargetFirmware GetTargetFirmware() {
```
SciresM, line 28: consteval?

libraries/libvapours/include/vapours/svc/board/generic/svc_device_name.hpp
```diff
@@ -0,0 +1,26 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#pragma once
+#include <vapours/svc/svc_types_common.hpp>
+
+namespace ams::svc::board::generic {
+
+    enum DeviceName {
+        /* If there is no IOMMU, there are no device names. */
```
SciresM, line 22: Recognizing that this is pedantic as hell: `smmu` in all lowercase instead of IOMMU in all uppercase?

mesosphere/kernel_ldr/source/kern_init_loader.cpp
```diff
@@ -87,6 +87,22 @@ namespace ams::kern::init::loader {
...
+        #ifdef ATMOSPHERE_BOARD_NINTENDO_NX
+
+        ALWAYS_INLINE bool ShouldPerformCpuSpecificSetup() {
+            /* Perform cpu-specific setup only on < 10.0.0. */
+            return kern::GetTargetFirmware() < ams::TargetFirmware_10_0_0;
+        }
+
+        #else
+
+        constexpr ALWAYS_INLINE bool ShouldPerformCpuSpecificSetup() {
```
SciresM, line 99: consteval?

libraries/config/arch/arm64/cpu/cortex_a57/cpu.mk
fincs: Actually I'm not sure. Arm designed processors can be configured with or without the extensions by system implementors. I guess technically it should be done on a board by board basis; but since the only A57 board we target does have the crypto extensions, for now this will do.

PR #1185, opened 2020-10-09
Add Github Workflow and Build Status

bladeoner:
- Adds a Github Workflow
- Shows the Build Status of the Github Workflow on the main page

You can view the Build Status button on my Github branch, the button will only be available when the commit has been merged and the build on has been completed:
https://github.com/bladeoner/Atmosphere/tree/workflow

At this moment I triggered a build which will successfully complete which can be seen here:
https://github.com/bladeoner/Atmosphere/actions

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -1,0 +1,53 @@
+name: Atmosphere Build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: ubuntu-20.04
+    runs-on: ubuntu-20.04
+    container: devkitpro/devkita64_devkitarm
+
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+
+    - name: Copy source list
+      run: sudo cp -r sources.list /etc/apt/
+      working-directory: ./workflow_files
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y dirmngr --install-recommends --allow-unauthenticated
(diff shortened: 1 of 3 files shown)


SciresM:
Atmosphere is uninterested in continuous integration/GitHub workflow at this time.

I plan to add continuous integration when Atmosphere is at version 1.0.0; this is expected to occur within the next year or so, as we reach the end of the "re-implement the entire boot image package" project. 

More to the point -- Atmosphere master is unstable, and I do not wish people in general to run builds of not-official-release commits. My life becomes drastically harder from a debugging point of view if this happens, as it greatly complicates the issue of determining what debug ELF to use when looking at fatal error reports.

GitHub workflow also seems problematic insofar as it relates to sept, which requires building with keys present only on my machine.

Finally, I do not use linux or ubuntu. I use Windows. 

I desire for whatever form of continuous integration we settle on to use Windows as well, so that it closely mimicks my own development environment. The use of e.g. ubuntu for this is undesirable to me.

In the future, I would strongly advise you reach out to me (or other members of the Atmosphere team like hexkyz or fincs) before putting in this kind of work, as it makes it less likely that stuff like this will be wasted time.

Best regards, and feel free to contact me on Discord to talk further (SciresM#0524).

SciresM:
I also notice, FYI, that your workflow appears to install and use python3.

Atmosphere only officially supports the usage of python2, and does not expect to ever transition to python3.

Any continuous integration solution will need to be sure to use python2 for building, to accommodate this.

bladeoner:
@SciresM thank you for your explanation.

Can you please update the dependencies in the documentation, to only use python2 because getting it to work with python3 was the most difficult part:
https://github.com/Atmosphere-NX/Atmosphere/blob/master/docs/building.md

Thanks in advance and enjoy your weekend.

PR #1201, opened 2020-11-01
Refactor boot sysmodule to be suitable for running on mariko

SciresM:
boot sysmodule has problems for mariko, we're fixing that


SciresM:
This is done pending validation/testing.

Console boots/seems to function totally normally on Erista using the current revision of the codebase, but given boot sysmodule does hardware configuration it merits (I think) substantial validation.

PR #1213, opened 2020-11-25
Correct all occurrences of "occured"

Lusamine:
occured
I noticed this occurring every time I crashed my Switch.  It would appear that "occurred" is the correct spelling.


SciresM:
revert the occur(r)ences which touch comments in library code, please.

Lusamine:
The comments should be restored to their naturally occurring state.

PR #1308, opened 2020-12-30
Add lz4 dependency to the build instructions

tuxuser:
At build-time of exosphere, the `split_program.py` script is called.
It requires python module `lz4` -> https://github.com/Atmosphere-NX/Atmosphere/blob/master/exosphere/program/split_program.py

Adding lz4 to the building.md documentation.


SciresM:
Remove "(Python library)" from dependency names, and I'll merge.

tuxuser:
Done

PR #1321, opened 2021-01-07
Fix PRODINFO blanking for newer versions.

CaramelDunes:
Following a discussion on Discord.

Original message:

My guess is that HOS is now checking Rsa2048DeviceCertificate and/or Rsa2048DeviceKey at boot time. Just like EccB233DeviceCertificate and ExtendedEccB233DeviceKey, a valid CRC is not enough i.e. the console's device id must match those of EccB233DeviceCertificate + 0xC6 and of the decrypted ExtendedEccB233DeviceKey. A simple solution would be to remove line 225 and 226 (https://github.com/Atmosphere-NX/Atmosphere/blob/c547c7f0e7a5de4292b58882f261be11b288ac0e/stratosphere/ams_mitm/source/amsmitm_prodinfo_utils.cpp#L225).
It should not affect the blanking capabilities as EccB233DeviceCertificate was not blanked in the first place (because you cannot simply zero it). Blanking the SSL cert is the important stuff here.
Incognito's implementation was probably created using the "blank as much stuff until it cannot boot anymore" strategy.

### DIFF
--- a/stratosphere/ams_mitm/source/amsmitm_prodinfo_utils.cpp
+++ b/stratosphere/ams_mitm/source/amsmitm_prodinfo_utils.cpp
@@ -222,8 +222,6 @@ namespace ams::mitm {
             Blank(info.GetBlock<EcqvEcdsaAmiiboRootCertificateBlock>());
             Blank(info.GetBlock<EcqvBlsAmiiboRootCertificateBlock>());
             Blank(info.GetBlock<ExtendedSslKeyBlock>());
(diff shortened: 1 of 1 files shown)


SciresM:
It looks like this PR's suggestion is to not blank data that we were previously blanking.

What is the RSA2048DeviceCertificate used for? What makes us confident it's not problematic to not blank that data any more?

SciresM:
Asked around, the data is "not important". Given the if, it wasn't being blanked on older units anyway....

...I'd like to say that I really don't understand people who want blanking, since this key supposedly contains the device ID, but whatever.

PR #1350, opened 2021-01-30
dns: add sfdnsres mitm

SciresM:
NOTE: PR currently not complete.

When complete, this PR should add an sfdnsres mitm, which allows redirecting user-specified hosts to user-specified ip addresses.

Current vague plan is sd:/atmosphere/hosts/defaults, and then sd:/hosts/emummc_<id>, sd:/hosts/sysmmc for overrides.


SciresM:
NOTE: PR is now theoretically functionally complete, pending review/sanity checks.

PR #1376, opened 2021-02-19
Add uart_mitm

yellows8:
This adds uart_mitm, for logging uart bluetooth HCI traffic.


misson20000:
The dynamic memory allocation is a little bit concerning to me. I'd suggest it be converted over to static allocation, but I suspect it's dynamic to avoid using the memory when the UART mitm is disabled. For that matter though, I do wonder if a uart mitm really belongs upstream? It seems kindof a niche debugging tool to me. If upstreamed at all, it probably belongs behind a compile-time enable.

yellows8:
The dynamic allocs are only done when uart.mitm is enabled via the setting.

PR #1379, opened 2021-02-23
docs/roadmap.md: Changed Word

Sebbl0508:


SciresM:
I'm not going to merge this. I'll think on it, though -- Mesosphere is a weird one, because it's not always-on yet. I might change that, but also change the date to be June 2021 or something.

PR #1406, opened 2021-03-19
daybreak: Added a warning when resetting to factory settings

Adubbz:
As per title.


SciresM:
Do you really not have the ability to commit directly to master?

PR #1407, opened 2021-03-19
Support DNS-MITM redirection exceptions

znxDomain:
Per a conversation on Discord you indicated you were amicable to such an enhancement, though you were uncertain about the prefix character of `-`.

Here is a first attempt at implementing this feature.  It works for me.


SciresM:
I am still thinking about how I want this to be implemented.

Please do not spend development time implementing features and then pull requesting them when I tell you things need to be thought through.

Rejecting this out of hand, in order to be consistent on that point.

PR #1421, opened 2021-04-08
Update dmnt_cheat_vm.cpp

easyworld:
* fix case fallthrough in function "LogOpcode"
* use same attribute name in one opcode case


SciresM:
Much obliged.

PR #1586, opened 2021-08-08
Add arguments support to Daybreak

HamletDuFromage:
Just a small change to be able to launch daybreak with a specified update path

### DIFF
--- a/troposphere/daybreak/source/main.cpp
+++ b/troposphere/daybreak/source/main.cpp
@@ -261,7 +261,10 @@ class Daybreak : public CApplication {
 
 int main(int argc, char **argv) {
     /* Initialize the menu. */
-    dbk::InitializeMenu(FramebufferWidth, FramebufferHeight);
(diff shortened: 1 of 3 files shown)


SciresM:
I don't understand -- hbmenu doesn't actually support arguments to daybreak like this?

Closing in the absence of an understanding why this is justified/desirable.

HamletDuFromage:
Well it'd be mostly for [envSetNextLoad()](https://switchbrew.github.io/libnx/env_8h.html#ab3fa9a2b9c3296e182c7caf618dbb4e0), though it also works with nxlink. Opening daybreak on the right folder from the get-go would be justifiable for this use case, and changes nothing for regular use

HamletDuFromage:
So, no chance of merging right?

SciresM:
I've given it some thought, and would accept a version of this modulo caveats, I think. Reach out to me on discord to discuss?

PR #1594, opened 2021-08-20
ams: replace sept with tsec firmware

SciresM:
This replaces sept with a custom tsec key derivation firmware.

NOTE: This does not use any TSEC exploits whatsoever; it is a well-signed TSEC binary assembled with envyas and signed with the real cauth key.

For more details, contact `SciresM#0524`.


SciresM:
This is a huge simplification/boot time speedup for fusee, done as part of prep for the upcoming fusee-cpp rewrite.

Merge pending testing in the following scenarios:
[ ] 1.0.0 Erista Switch
[ ] < 6.2.0 Erista Switch
[ ] 6.2.0 Erista Switch

We also probably want to coordinate merge with hekate.

SciresM:
NOTE: We shouldn't do a release until hekate is ready, to avoid breaking shit.

PR #933, opened 2020-04-27
Reimplement LogManager sysmodule

XorTroll:
Reimplements LogManager (LogManager.Prod in retail) sysmodule, redirecting log data to log files in the SD card.
There isn't much left to do (everything seems to work fine), it just needs testing and a proper review.

PR #1617, opened 2021-09-12
LogManager: implement system module, client api, logging api

SciresM:
This supersedes #933.

LogManager (/logging) is a necessary component of tma2, got this done this week in order to be able to resume focus on tma2 development.


SciresM:
Some notes:

* Unless `atmosphere!enable_log_manager` is true, Nintendo's log manager will be used instead.
  * This prevents paying memory costs for LM when not enabling logging.
  * To facilitate this, Atmosphere's log manager has a different program id from Nintendo's.
  * `atmosphere!enable_htc` implies `atmosphere!enable_log_manager`.
* LogManager logs to tma, and the SD card (if `lm!enable_sd_card_logging` is true, which it is by default).
* Binary logs are saved to `lm!sd_card_log_output_directory`, which is `atmosphere/binlogs` by default.

PR #1662, opened 2021-10-10
docs: add hactool to list of build dependencies

ndeadly:


SciresM:
Kind of a weird one, because emummc depends on it, not ams per se.

PR #1688, opened 2021-10-31
ams: Allow standalone gdbstub while starlink is in development.

SciresM:
This adds experimental support for standalone gdbstub (dmnt.gen2), by setting `atmosphere!enable_standalone_gdbstub = u8!0x1` with enable_htc as u8!0x0.

This hosts atmosphere's gdbstub as a standard socket on port 22225.

I'm thinking it makes sense to merge this and include it in releases (certainly until starlink is readily available/usable), but it would be nice to have a higher degree of certainty that our gdbstub is featureful and useful for people's use cases.

single-step/watchpoints should both be functional.

Notable comments:
* While this is enabled, cheats won't work, because the cheat engine is not yet integrated into dmnt.gen2. This is a TODO, it's hard to cleanly integrate because the cheat engine code is pretty terrible in many ways.
* Determining memory layout needs some work. Not sure what stub command we need to add to make this nice and easy? 
* I think loading symbols from PC/ELF doesn't work quite right, not 100% sure what needs to be done for this either.
* I am interested in custom gdb remote commands to e.g. allow auto-attaching to next application or to next homebrew NRO or to specific program ID.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,7 @@ dist-no-debug: all
 	mkdir -p atmosphere-$(AMSVER)/stratosphere_romfs/atmosphere/contents/0100000000000042
 	mkdir -p atmosphere-$(AMSVER)/stratosphere_romfs/atmosphere/contents/0100000000000420
 	mkdir -p atmosphere-$(AMSVER)/stratosphere_romfs/atmosphere/contents/010000000000B240
+	mkdir -p atmosphere-$(AMSVER)/stratosphere_romfs/atmosphere/contents/010000000000D609
 	mkdir -p atmosphere-$(AMSVER)/stratosphere_romfs/atmosphere/contents/010000000000D623
 	cp stratosphere/boot2/boot2.nsp atmosphere-$(AMSVER)/stratosphere_romfs/atmosphere/contents/0100000000000008/exefs.nsp
 	cp stratosphere/dmnt/dmnt.nsp atmosphere-$(AMSVER)/stratosphere_romfs/atmosphere/contents/010000000000000D/exefs.nsp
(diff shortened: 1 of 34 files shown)


SciresM:
Test build for this. Don't use if you don't know what you're doing, but if you do I welcome your thoughts, feedback, and possibly your contributions. 

Contact `SciresM#0524` as needed. 

[atmosphere-1.2.2-allow_standalone_gdbstub-90473a230.zip](https://github.com/Atmosphere-NX/Atmosphere/files/7449928/atmosphere-1.2.2-allow_standalone_gdbstub-90473a230.zip)

SciresM:
New build with the latest changes, includes nicer "monitor get info", fix for crash on unknown monitor command, functional ctrl-c break/suspend.

[atmosphere-1.2.2-allow_standalone_gdbstub-d2b024f19.zip](https://github.com/Atmosphere-NX/Atmosphere/files/7455622/atmosphere-1.2.2-allow_standalone_gdbstub-d2b024f19.zip)

SciresM:
New build with latest changes.

`monitor get mapping <address>`
`monitor get mappings`

Doubled the packet buffer size (fake fix for issues with lots of threads/mappings).

Added support for detecting thread names. Gets real name of Nintendo and libstratosphere threads.

Made it so that if you select a thread and then hit ctrl-c, then if that thread is still alive you see that thread as being interrupted instead of the most recent thread to receive an event.

[atmosphere-1.2.2-allow_standalone_gdbstub-ea61ac9ea.zip](https://github.com/Atmosphere-NX/Atmosphere/files/7456887/atmosphere-1.2.2-allow_standalone_gdbstub-ea61ac9ea.zip)

SciresM:
New build with latest changes.

* We shouldn't write out of bounds when generating packets any more.
* Monitor get mappings now accepts optional address to start from, and tells you what address to continue from when there are too many mappings to do them in one go. (TODO: Is it possible to return all the mappings magically despite limited buffer size? Send multiple packets? idk).
* Fixed thread name_(nil).
[atmosphere-1.2.2-allow_standalone_gdbstub-545ade835.zip](https://github.com/Atmosphere-NX/Atmosphere/files/7462762/atmosphere-1.2.2-allow_standalone_gdbstub-545ade835.zip)

SciresM:
I am going to merge this, so that it can be included in ams 1.2.3.

It's definitely still experimental, but hey I'm sure we'll get way more testers this way, lol.

PR #1700, opened 2021-11-06
Minor header fixes to reduce parsing issues with Clang

leoetlino:
A small set of fixes to libstratosphere headers to make stratosphere.hpp more parsable by Clang.

At the time of writing Clang has incomplete C++20 support so two of the commits add small, non-intrusive workarounds to allow Clang-based tools to still parse most of stratosphere.hpp and dump type definitions.

Unfortunately getting to 0 compilation errors is not possible at this time because Clang appears to have bugs in its constant evaluation engine and its definition of literal types does not conform to the C++17 standard. (That is the reason why the binary-search based offsetof implementation is broken in some cases on Clang.)

This PR is not just Clang workarounds though, it also fixes some non-standards-compliant code :)

### DIFF
--- a/libraries/libstratosphere/include/stratosphere/sm/sm_types.hpp
+++ b/libraries/libstratosphere/include/stratosphere/sm/sm_types.hpp
@@ -40,8 +40,8 @@ namespace ams::sm {
             return out;
         }
 
-        static constexpr ServiceName Encode(const char *name) {
-            return Encode(name, std::strlen(name));
+        static constexpr ServiceName Encode(util::string_view name) {
+            return Encode(name.data(), name.size());
         }
     };
 
--- a/libraries/libvapours/include/vapours/results/results_common.hpp
+++ b/libraries/libvapours/include/vapours/results/results_common.hpp
@@ -177,7 +177,11 @@ namespace ams {
                 static_assert(Value != Base::SuccessValue, "Value != Base::SuccessValue");
             public:
                 constexpr ALWAYS_INLINE operator Result() const { return MakeResult(Value); }
-                constexpr ALWAYS_INLINE operator ResultSuccess() const { OnResultAbort(Value); }
+                constexpr ALWAYS_INLINE operator ResultSuccess() const {
+                    OnResultAbort(Value);
+                    __builtin_unreachable();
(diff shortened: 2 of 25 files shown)


libraries/libvapours/include/vapours/results/results_common.hpp
```diff
@@ -177,7 +177,10 @@ namespace ams {
                 static_assert(Value != Base::SuccessValue, "Value != Base::SuccessValue");
             public:
                 constexpr ALWAYS_INLINE operator Result() const { return MakeResult(Value); }
-                constexpr ALWAYS_INLINE operator ResultSuccess() const { OnResultAbort(Value); }
+                constexpr ALWAYS_INLINE operator ResultSuccess() const {
+                    OnResultAbort(Value);
+                    return ResultSuccess();
```
SciresM, line 182: I am very uncomfortable with this. OnResultAbort(Value) is NORETURN... Can __builtin_unreachable() or AMS_ASSUME(false) be used instead?

libraries/libstratosphere/include/stratosphere/sm/sm_types.hpp
```diff
@@ -40,8 +40,8 @@ namespace ams::sm {
             return out;
         }
 
-        static constexpr ServiceName Encode(const char *name) {
-            return Encode(name, std::strlen(name));
```
SciresM, line 44: Wonder if we should just add e.g. util::Strlen. We have (as Nintendo does) Strnlen, Strlcpy...ah well, this is probably fine.

SciresM: (no need to make changes here).

SciresM (changes requested):
Pretty much just one change needed.

There's some minor style stuff too but I think it's all fine, nothing we don't already have in other places in the codebase.

libraries/libvapours/include/vapours/results/results_common.hpp
leoetlino: Added __builtin_unreachable() before the return. Unfortunately Clang requires a return statement to be present here even if the end of the function can never be reached...

PR #1724, opened 2021-11-29
Expand Max Cheat Count

proferabg:
Expand Max Cheat Count to 1024 instead of 128.

Maybe add in config option if this isn't good enough or not suitable for everyone?


SciresM:
No, the memory cost is unacceptable.

Please reach out to me on discord to discuss before just making a PR like this.

masagrator:
And 0x1000 is 4096, not 1024. 😅

proferabg:
> And 0x1000 is 4096, not 1024. 😅

I realized this after I made the commit. It was discussed that this isn't the way to go about it.

PR #1780, opened 2022-02-14
Integrate new result macros.

SciresM:
Going to leave this up for a few hours (while I review it) before merge.

This adds some new result macros -- functions which use them (and don't use normal result statements) may make use of ON_RESULT_FAILURE and other new "magic" return-introspection macros.

I want to get this merged before the codebase diverges more than it has to.


SciresM:
NOTE: This does not use the new result macros universally; code which does not use them remains valid, and so not everything is converted.

PR #1788, opened 2022-03-06
KThreadContext: Use El0PsrMask constant in GetUserContext

merryhime:
Minor nit.


SciresM:
lol, did I define this and not use it

SciresM:
Thank you very much :)

PR #1789, opened 2022-03-06
ams: support building unit test programs on windows/linux/macos

SciresM:
**NOTE: This does not enable running official software, and is not intended to provide any functionality relating to official software in any way. This is wholly unit testing/utility program machinery.**

This adds support for building libstratosphere (and then linking/using it on native programs) for the following targets:
* Windows x64
* Linux x64
* macOS x64
* Linux arm64
* macOS arm64

A test program ("TestFs") is included, and verifies the correct result codes are returned by the FS client apis for all supported operations on native paths.

To facilitate this, the following was done:
* AMS's build system was rewritten, and is now saner. We now support building release/debug/assert builds for all supported targets (NX, win_\*, macos_\*, linux_\*). Some longstanding dependency issues were also fixed.
* fs:: client library was refactored enormously to both reflect latest (13.x) APIs/logic, and to support properly a sf::SharedPointer based object for FS interaction; on NX, this is still using libnx primitives, but the logic can be reused verbatim for future FS work.
* Many changes were made to support compiling with clang; note that clang-14 is required for this to work.
* Many things which were previously arm64 only now have x64 logic.

The following build environments were tested:
* Windows x64, building with MinGW. Note that MinGW is *assumed* when building for windows. Clang support may come at a future date, MSVC probably never, unless they seriously step up their constexpr game.
* Linux x64, building with either gcc or clang. If building with gcc, libstdc++ is used. If building with clang, libc++ is used.
* macOS x64, building with clang. 
* Linux arm64, building with clang. Building with g++ is intended for the future.
* macOS arm64, building with clang. Tested using self-build clang-master from git on a 2020 M1 Macbook Air.

Build instructions not currently provided; if you are interested in helping write unit tests, DM `SciresM#0524` on Discord and I can help get you set up, probably.

Automated building/running of tests is still TODO. Also TODO is cross-compiling; right now we assume that you are natively building windows x64 from windows, linux from linux, macOS from macOS.

Note that libjpeg-turbo is required, as usual. On macOS/linux, /opt/libjpeg-turbo is added to include path (for now), as that is where the official release builds of libjpeg-turbo (from [here](https://libjpeg-turbo.org/DeveloperInfo/PreReleases)) install to.

Also note that any and all of this is subject to change.

This PR is not intended for immediate merging, but I wanted to get it in a PR-able state so that if atmosphere needs to do a release, it can be merged/we don't end up with ridiculously divergent logic.

Windows implementations based on reverse engineering of unity logic shipped with commercial games. Linux/macOS logic based on reading man pages found by googling, and by pestering @fincs and @mtheall with a lot of questions.


libraries/libstratosphere/source/fssystem/fssystem_local_file_system.cpp
```diff
@@ -0,0 +1,1751 @@
...
+                                } else {
+                                    out_entry = std::addressof(m_temp_entries[m_temp_entries_ofs++]);
+                                }
+
+                                /* Setup the output entry. */
+                                {
+                                    std::memset(out_entry->name, 0, sizeof(out_entry->name));
+
+                                    const auto name_len = std::strlen(ent->d_name);
+                                    std::memcpy(out_entry->name, ent->d_name, name_len + 1);
```
misson20000, line 749: Uh, bounds check here? Or otherwise leave a comment/static assert here explaining why it's not necessary.

libraries/libstratosphere/source/fssystem/fssystem_local_file_system.cpp
```diff
@@ -0,0 +1,1751 @@
...
+                        R_UNLESS(attr == INVALID_FILE_ATTRIBUTES, fs::ResultPathAlreadyExists());
+
+                        /* Return the original result. */
+                        R_THROW(R_CURRENT_RESULT);
+                    }
+                } R_END_TRY_CATCH;
+            }
+            #else
+            {
+                /* ::rename() will destroy an existing file at new path...so check for that case ahead of time. */
```
misson20000, line 1445: `renameat2` has a flag to error out if the destination exists (`RENAME_NOREPLACE`)

libraries/libstratosphere/source/fssystem/fssystem_local_file_system.cpp
SciresM: I don't think it's necessary because of path max being less than fs::EntryNameLengthMax on both linux and macOS? I can add an error check though, if you think it would be wise.

libraries/libstratosphere/source/fssystem/fssystem_local_file_system.cpp
SciresM: I could add this, but note that I would have to ifdef because macOS does not have `renameat2`.

libraries/libstratosphere/source/fssystem/fssystem_local_file_system.cpp
misson20000: Yeah, that's what I figured, so a nearby `static_assert(PATH_MAX <= fs::EntryNameLengthMax)` would keep anyone like me from asking in the future.

libraries/libstratosphere/include/stratosphere/fs/fs_read_only_filesystem.hpp
```diff
@@ -115,57 +115,57 @@ namespace ams::fs {
...
             virtual Result DoGetTotalSpaceSize(s64 *out, const fs::Path &path) override final {
                 AMS_UNUSED(out, path);
-                return fs::ResultUnsupportedOperationInReadOnlyFileSystemTemplateB();
+                return fs::ResultUnsupportedCommitProvisionallyForReadOnlyFileSystem();
             }
 
             virtual Result DoCommitProvisionally(s64 counter) override final {
                 AMS_UNUSED(counter);
-                return fs::ResultUnsupportedOperationInReadOnlyFileSystemTemplateC();
+                return fs::ResultUnsupportedGetTotalSpaceSizeForReadOnlyFileSystem();
```
Morph1984, line 168: In case it was missed, these are reversed.

SciresM: I'll need to check if these are wrong names or if these are wrong values.

SciresM:
Merging this pre-maturely, so that I can begin work on 14.0.0.

Work will resume on this post-14.0.0.

PR #1814, opened 2022-03-25
build: don't choke on stratosphere.hpp.gch file from unclean pre-1.3.0 build dir

justinkb:
on <=1.2.6 builds, stratosphere.hpp.gch will be a file. The makefile doesn't account for this possibility.

this obviously leads to a failure to mkdir this path


SciresM:
This was kind of intentional. Not sure that I like a general case file deletion there. Vs just telling people to true reclone.

justinkb:
ah! I did not consider that. feel free to close entirely :-) maybe add a small line in the 1.3.0 release notes?

PR #1924, opened 2022-07-25
ams_mitm: allow displaying HTML content in applications without a manual NCA

tech-ticks:
Always allows loading HTML from `sd:/atmosphere/contents/<application_id>/manual_html` if the directory exists with at least one file or folder, even if the game doesn't have a manual NCA.


SciresM:
There are a lot of little style things here -- I'm going to ignore this PR and go implement this myself, since it's probably faster than explaining in exacting detail how this change should be made.

No offense intended, feel free to talk to me on discord about it.

PR #1981, opened 2022-11-04
Update ams_api_version.h

arch-box:
No issues experienced on HOS version 15.0.1 with Atmosphere release 1.4.0


SciresM:
You don't need to PR this kind of thing. It's technically correct, but I like to increment that when I update other API versions.

PR #2163, opened 2023-06-26
Add automated CI builds with GitHub Actions

hummeltech:
This pull request adds automated CI builds using GitHub Actions.

I noticed that there were at least two closed pull requests pre-`1.0.0` adding continuous integration builds and since Atmosphère `1.0.0` has been released (which was one of the stipulations of adding continuous integration builds) and no additional pull requests adding continuous integration builds have been created, I decided to give it another try.

Getting this to also build using `Windows` as suggested would be tricky as it seems most (if not all) of the IPs used by GitHub Actions have been banned from using `pkg.devkitpro.org` and devkitPro's recommendation is to just use the docker images they build and upload to `hub.docker.com`. Unfortunately they do not provide a Windows docker image which would likely solve most (if not all) of this issue.

The aforementioned IP banning is the reason why `hactool` is needing to be built here, luckily it only takes about 10-20 seconds.

Feel free to let me know if there are any comments/suggestions or if it would be preferable to just close this pull request.

Thanks and have a nice day!

Here is an example of a successful Workflow run:
https://github.com/hummeltech/Atmosphere/actions/runs/5373884521/jobs/9748692838

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,64 @@
+---
+name: Build
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  Linux:
+    container:
+      image: devkitpro/devkita64:latest
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Install dependencies
+        run: |
+          apt-get --yes update
+          apt-get --yes install build-essential python-pip zstd
+
+      - name: Install `python` dependencies
+        run: |
+          pip install lz4 pycryptodome
+
+      - name: Build & Install `hactool`
+        run: |
+          echo 'MAKEFLAGS="-j'$(($(nproc) + 1))'"' >> /opt/devkitpro/pacman/etc/makepkg.conf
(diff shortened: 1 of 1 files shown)


SciresM:
No, sorry.

SciresM:
Please reach out to me on discord (sciresm) before doing this kind of work...I have very strong feelings on what kinds of CI I would accept, and especially that "Windows" is the build platform that matters.

In addition, I do not actually want normal users using random commits, as I don't make an effort to make all commits build or safe to use.

Also please note that I am currently on travel without access to a computer and have intermittent internet access until July 15th, and may be much slower than usual about replying to anything.

hummeltech:
Hello @SciresM, I don't use (and will never use) Discord so this will have to be the place for these discussions to occur. Also, I don't mind waiting for you to return, I could get Windows builds to work as well, if that is the build platform that matters.

Good point about random users running random commits, the simplest solution that that is to not upload any non-report artifacts which would make this serve to only verify that the build (as well as any existing tests) complete successfully.

SciresM:
Well, I am unwilling to merge PRs without some kind of one on one real-time conversation occurring ahead of time, so if you don't use discord you may just be literally unable to contribute to Atmosphere.

PR #2201, opened 2023-10-12
Basic ("the console boots and functions as expected") 17.0.0 support

SciresM:
This implements such updates as required to boot 17.0.0. It also implements a workaround for a change in FS that would cause consoles which have used emmc rebuild in the past to brick on update-to-17.0.0.


TudouHacks:
W

PR #2248, opened 2023-12-06
exo: Invert quest state.

laleeroy:
Can you take a look at this?
Seems like it bypass the error that shows upon booting on  demo/kiosk units


SciresM:
This would cause problems on a non-kiosk unit. Why on earth would I merge this?

Issue #934, opened 2020-04-28
Add Support For Float/Double Cheats


rydoginator:
## Feature Request

I am requesting to add float/double casting to the dmnt_cheat process. 

#### Overview:
- Currently the supported types are 8 bit,16 bit, 32 bit and 64 bit. Which is nice, but if you're working with a value in memory that is a float or double, then the arithmetic instructions won't work evenly. 

#### Smaller Details: 
- Adding F and D to determine the size of the value, like how 1,2,4,8 are already used.
#### Nature of Request:  
  - Addition of more data types for sizes of cheats. 
 - Looks like the size is determined [here](https://github.com/Atmosphere-NX/Atmosphere/blob/master/stratosphere/dmnt/source/cheat/impl/dmnt_cheat_vm.cpp#L317), so it would be a matter of modifying that switch-case statement.  
I just can't for the life of me figure out how to compile, I get errors all over the place, which is why I'm asking as a feature request :)

### What component do you feel this would best fit within?
- [Stratosphère](https://github.com/Atmosphere-NX/Atmosphere#components)</br>
    - The dmnt_cheat_vm custom sys-module.</br>

### Why would this feature be useful?
Like I said, implementing float/double types will allow for more better cheats, like moon jump or size modifiers, basically anything that modifies anything that is float or double. </br>

masagrator:
But those are Hex values, not decimal. So for example 30 float you are putting as
```
04000000 01234567 41F00000
```

rydoginator:
Yes, I know what floats would be in hex. 
I'm mainly concerned about arithmetics. 
For example, if you were to add 3F800000 + 3F800000 together, you'd get 7F000000 which is 1.70141e+38 in float. Casting it as a float would cause them to be added evenly so that it would be 40000000 which would be 2.0.

SciresM:
Fine with adding this once C++20 std::bit_cast is available.

rydoginator:
> Fine with adding this once C++20 std::bit_cast is available.

Why not a standard cast?

SciresM:
how do you "standard cast" a hex pattern to float/double

rydoginator:
> how do you "standard cast" a hex pattern to float/double

[...]

PR #2479, opened 2025-01-05
dmnt_extension

tomvita:
1. Extension to Type 0
	Purpose: Adding the "none" option to improve flexibility, especially when game code resides in a module. This allows the setup of register R such that R corresponds to the module's main address.
		M: Memory region to write to (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr, 4 = none).

2. Extension to type 5
	Purpose: Introducing two new extensions for enhanced pointer chain manipulation.
	Extension 2
	Allows separate source and target registers, removing the need to save the source register for use in other cheats that share the initial part of a pointer chain.
		#### Load from Register Address Encoding
		`5T0R2SAA AAAAAAAA`
		+ T: Width of memory read (1, 2, 4, or 8 bytes).
		+ R: Register to load value into. 
		+ S: Register to use as the base memory address.
		+ A: Immediate offset to use from register R.
	Extension 3
	Useful for cases where game code resides in a module. It allows the setup of register R such that main + R points to the module’s main address.
		#### Load From Fixed Address Encoding with offset register
		`5TMR3SAA AAAAAAAA`
		+ T: Width of memory read (1, 2, 4, or 8 bytes).
		+ M: Memory region to write to (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr).
		+ R: Register to load value into.
		+ S: Register to use as offset register.
		+ A: Immediate offset to use from memory region base.

3. Extension to type 8
	Purpose: Enhancing keypad values to support cheats that execute only once per key press, making them ideal for actions like incrementing or decrementing specific values.
		8000000: when this is set button only activate code once per keydown, need to be release before the code will run again

4. Extension to type 9
	Purpose: Adding support for new arithmetic operations to handle floating-point and double-precision calculations.
		New Arithmetic Types:
		+ 10: Float Addition, Width force to 4 bytes
		+ 11: Float Multiplication, Width force to 4 bytes
		+ 12: Double Addition, Width force to 8 bytes
		+ 13: Double Multiplication, Width force to 8 bytes

5. Extension to type C0
	Purpose: Adding extension 6 to validate a hook before applying hacks. This is particularly useful for games with multiple loadable modules, where the build ID alone may not prevent hacking the wrong code.
		C0Tcr6Ma aaaaaaaa VVVVVVVV (VVVVVVVV)
		6: Compare [Memory Base + Offset Register + Relative Offset] against Static Value

### DIFF
--- a/docs/features/cheats.md
+++ b/docs/features/cheats.md
@@ -62,11 +62,13 @@ Code type 0x1 performs a comparison of the contents of memory to a static value.
 If the condition is not met, all instructions until the appropriate End or Else conditional block terminator are skipped.
 
 #### Encoding
-`1TMC00AA AAAAAAAA VVVVVVVV (VVVVVVVV)`
+`1TMCXrAA AAAAAAAA VVVVVVVV (VVVVVVVV)`
 
-+ T: Width of memory write (1, 2, 4, or 8 bytes).
-+ M: Memory region to write to (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr).
++ T: Width of memory read (1, 2, 4, or 8 bytes).
++ M: Memory region to read from (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr, 4 = none).
 + C: Condition to use, see below.
++ X: Operand Type, see below.
++ r: Offset Register (operand types 1).
 + A: Immediate offset to use from memory region base.
 + V: Value to compare to.
 
@@ -78,6 +80,9 @@ If the condition is not met, all instructions until the appropriate End or Else 
 + 5: ==
 + 6: !=
 
+#### Operand Type
++ 0: Memory Base + Relative Offset
++ 1: Memory Base + Offset Register + Relative Offset
 ---
 
 ### Code Type 0x2: End Conditional Block
@@ -126,7 +131,7 @@ Code type 0x5 allows loading a value from memory into a register, either using a
 `5TMR00AA AAAAAAAA`
 
 + T: Width of memory read (1, 2, 4, or 8 bytes).
-+ M: Memory region to write to (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr).
++ M: Memory region to write to (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr, 4 = none).
 + R: Register to load value into.
 + A: Immediate offset to use from memory region base.
 
@@ -149,7 +154,7 @@ Code type 0x5 allows loading a value from memory into a register, either using a
 `5TMR3SAA AAAAAAAA`
 
 + T: Width of memory read (1, 2, 4, or 8 bytes).
-+ M: Memory region to write to (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr).
++ M: Memory region to write to (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr, 4 = none).
 + R: Register to load value into.
 + S: Register to use as offset register.
 + A: Immediate offset to use from memory region base.
--- a/docs/features/cheats.md
+++ b/docs/features/cheats.md
@@ -200,7 +200,7 @@ Code type 0x8 enters or skips a conditional block based on whether a key combina
 
 + k: Keypad mask to check against, see below.
 
-Note that for multiple button combinations, the bitmasks should be OR'd together. Also, bit 27 does not correspond to a button and instead acts as a no auto-repeat flag.
+Note that for multiple button combinations, the bitmasks should be ORd together.
 
 #### Keypad Values
 Note: This is the direct output of `hidKeysDown()`.
@@ -231,7 +231,6 @@ Note: This is the direct output of `hidKeysDown()`.
 + 0800000: Right Stick Down
 + 1000000: SL
 + 2000000: SR
-+ 8000000: No auto-repeat. When this bit is set, the conditional block executes only once when the keypad mask matches. The mask must stop matching to reset for the next trigger.
 
 ---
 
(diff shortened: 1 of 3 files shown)


SciresM:
You should have reached out to me on discord before doing this work.

Most of this is fine, and I have no real objection to most of it, but e.g. I am not sure I would accept the float/double stuff as-is, I'm not really sure that I understand the C0 extension, and I am not sure I'm sold on the keypress-once thing. Does bitmask `8000000` ever actually exist as hid output? I'm not fully convinced that it doesn't, these days, though maybe it does not.

Also, the code for the keypress-once thing violates atmosphere style, and the docs have bad English. Please reach out to me on Discord, if you want this to make it in. All that said, there *is* a fair amount of stuff here that's not super objectionable and which I'd allow to be merged.

SciresM:
Yeah, that key press thing will need to work a different way. I would do it as a separate code. That bit (0x08000000) corresponds to a real button press (right-SR): https://github.com/switchbrew/libnx/blob/master/nx/include/switch/services/hid.h#L284

Generally, this stuff should be a new code, it only works the way that it does because we had to do legacy support for the original gateway cheat vm.

tomvita:
I believe the following would work with the current implementation, but it is an undocumented feature since it's not included in the current documentation:

    4000000: SL Right Joy-Con
    8000000: SR Right Joy-Con

Is it be a problem to change to 

   4000000: SL Right Joy-Con
   8000000: No auto-repeat. When this bit is set, the conditional block executes only once when the keypad mask matches. The 
                   mask must stop matching to reset for the next trigger.

SciresM:
"I believe the following would work with the current implementation, but it is an undocumented feature since it's not included in the current documentation:"

It *is* documented, because the documentation says it uses the output of hidKeysDown() directly.

It will need to be a separate new command, if you want it in. And generally, I think that we should have a better key-press command, because the current one doesn't support all button presses.

tomvita:
I have removed the type 08 extension and make instead type c4 extension.
I have removed the type C0 extension and make instead type 01 extension.

docs/features/cheats.md
```diff
@@ -62,11 +62,13 @@ Code type 0x1 performs a comparison of the contents of memory to a static value.
 If the condition is not met, all instructions until the appropriate End or Else conditional block terminator are skipped.
 
 #### Encoding
-`1TMC00AA AAAAAAAA VVVVVVVV (VVVVVVVV)`
+`1TMCXrAA AAAAAAAA VVVVVVVV (VVVVVVVV)`
 
-+ T: Width of memory write (1, 2, 4, or 8 bytes).
-+ M: Memory region to write to (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr).
++ T: Width of memory read (1, 2, 4, or 8 bytes).
++ M: Memory region to read from (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr, 4 = none).
```
SciresM, line 68: Change this (and other cases) to "non-relative"

docs/features/cheats.md
```diff
@@ -126,7 +131,7 @@ Code type 0x5 allows loading a value from memory into a register, either using a
 `5TMR00AA AAAAAAAA`
 
 + T: Width of memory read (1, 2, 4, or 8 bytes).
-+ M: Memory region to write to (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr).
++ M: Memory region to write to (0 = Main NSO, 1 = Heap, 2 = Alias, 3 = Aslr, 4 = none).
```
SciresM, line 134: "non-relative"

docs/features/cheats.md
```diff
@@ -380,6 +404,61 @@ Code type 0xC3 reads or writes a static register with a given register.
 
 ---
 
+### Code Type 0xC4: Begin Extended Keypress Conditional Block
+Code type 0xC4 enters or skips a conditional block based on whether a key combination is pressed.
+
+#### Encoding
+`C4r000kk kkkkkkkk`
```
SciresM, line 411: Is it acceptable to make this `C4r00000 kkkkkkkk kkkkkkkk`?

stratosphere/dmnt/source/cheat/impl/dmnt_cheat_vm.cpp
```diff
@@ -769,6 +790,7 @@ namespace ams::dmnt::cheat::impl {
         return true;
     }
 
+    static u64 keyold = 0;
```
SciresM, line 793: static variables should be `s_variable_name`

stratosphere/dmnt/source/cheat/impl/dmnt_cheat_vm.cpp
```diff
@@ -1022,6 +1060,30 @@ namespace ams::dmnt::cheat::impl {
                             case RegisterArithmeticType_None:
                                 res_val = operand_1_value;
                                 break;
+                            case RegisterArithmeticType_FloatAddition:
+                                if (cur_opcode.perform_math_reg.bit_width == 4)
```
SciresM, line 1064: atmosphere style *always* uses braces, so this should be 
```
if (...) {
   // ...
} else if (...) { 
  // ...
}
```

Also, do not do direct casting like this. 1) atmosphere disallows the use of `&` as addressof operator, always use std::addressof.

2) This should be using std::bit_cast<>, not point logic. Please fix.

stratosphere/dmnt/source/cheat/impl/dmnt_cheat_vm.hpp
```diff
@@ -59,6 +60,7 @@ namespace ams::dmnt::cheat::impl {
         MemoryAccessType_Heap    = 1,
         MemoryAccessType_Alias   = 2,
         MemoryAccessType_Aslr    = 3,
+        MemoryAccessType_Blank   = 4,
```
SciresM, line 63: Change this to "_NonRelative"

SciresM (changes requested):
Some naming/style changes, also would be nice for C4.... to take two DWORDs as future-proof.

docs/features/cheats.md
tomvita: fixed

stratosphere/dmnt/source/cheat/impl/dmnt_cheat_vm.cpp
tomvita: fixed

stratosphere/dmnt/source/cheat/impl/dmnt_cheat_vm.hpp
tomvita: fixed

stratosphere/dmnt/source/cheat/impl/dmnt_cheat_vm.cpp
tomvita: changed to s_keyold

docs/features/cheats.md
tomvita: changed accordingly

tomvita:
updated according to review

tomvita:
## Enhancements and New Features

### Enhanced Memory Addressing
Improved capabilities for the following code types:
- `0x0`: Store Static Value to Memory
- `0x1`: Begin Conditional Block
- `0x5`: Load Register with Memory Value

### Floating-Point Arithmetic Support
Expanded functionality of code type `0x9` (*Perform Arithmetic*) to include operations with floating-point values.

### New Code Type - `0xC4`
Added code type `0xC4` (*Begin Extended Keypress Conditional Block*), enabling:
- Distinction between **key held** and **key down** states.
- Access to the full output of `hidKeysDown()` for enhanced input handling.

SciresM:
okay, this seems fine. I'll merge it if you're happy with it?

SciresM:
I suppose this would close #934.

tomvita:
Thank you very much.

PR #2544, opened 2025-05-27
Clarify that code type 0xC4 uses a 64-bit value in cheats.md by forma…

M1CR0H4CK3R:
…tting the key values as such

Changes the formatting of the key values at `Code Type 0xC4: Begin Extended Keypress Conditional Block` in cheats.md so that it is perfectly clear that those values are 64-bit values rather than 32-bit or 28-bit like `Code Type 0x8: Begin Keypress Conditional Block`, and also for the fact that the formatting matches the rest of the document and is thus cleaner.


SciresM:
In the future, please contact me on discord before opening pull requests @M1CR0H4CK3R

PR #2556, opened 2025-06-03
loader: add usb 3.0 enable patches for 20.1.0

lsp199308:


SciresM:
Noting that I am on vacation internationally until June 15th; correspondingly, this will not make it into an ams release for some time. Thank you for updating these, since I somehow missed that usb was updated.

PR #2603, opened 2025-08-08
Warning fixes

comex:
- nanovg: disable some compiler warnings

  Disable all warnings seen on devkitA64 GCC 15.1.0:

  - Several false-positive `-Wmisleading-indentation` warnings (the
    indentation is not actually misleading)

  - One `-Wuse-after-free` warning, which looks alarming and represents
    very dubious design, but is ultimately harmless (a freed pointer is
    passed to a function that doesn't use it).

  - Some `-Wunused-function` warnings in `simd/neon.h`

- daybreak: fix strncpy compiler warning

  GCC warned that the output string would not be nul-terminated if the
  input string was too long.  Fix this by subtracting 1 from the size
  argument, as is done for other `strncpy` calls in the file.  (It would
  probably be better to avoid `strncpy` entirely, but this is just a
  simple fix.)

- haze: fix duplicate-macro warnings

  haze was including both `<switch.h>` and `vapours/types.hpp` which both
  define `R_SUCCEEDED` and `R_FAILED`, producing compiler warnings.

  The intent is that `vapours/types.hpp` only gets included when targeting
  not-Switch and/or not-EL0.  But the check didn't account for
  Troposphère.  Fix that.

### DIFF
--- a/troposphere/daybreak/nanovg/Makefile
+++ b/troposphere/daybreak/nanovg/Makefile
@@ -50,6 +50,8 @@ ARCH	:=	-march=armv8-a+crc+crypto -mtune=cortex-a57 -mtp=soft -fPIE
 CFLAGS	:=	-g -Wall -O2 -ffunction-sections \
 			$(ARCH) $(DEFINES)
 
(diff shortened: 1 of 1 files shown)


SciresM:
Much obliged.

PR #2607, opened 2025-08-16
build: portability fixes

comex:
- Invoke Python scripts with `python3` if `python` is not available.

  Despite SciresM preferring Python 2, the scripts used in the build all
  worked in Python 3 already, so this is just a Makefile change.

- Export `MAKE` to work around a GNU bug.

  See the comment for details.

- `rmdir --ignore-fail-on-non-empty` -> `rmdir 2>/dev/null`

  macOS `rmdir` doesn't support `--ignore-fail-on-non-empty`.

  This is a slight downgrade in functionality, because now we ignore all
  errors from `rmdir` rather than just 'Directory not empty'.  I could
  have avoided this with a more complicated fix, but the benefit strikes
  me as not worth the complexity.  Let me know if you disagree.

- Append `$(DEVKITPRO)/tools/bin/` to `$PATH` when invoking
  `build_romfs`.

  This avoids the need to add devkitPro directories to `$PATH` when
  building Atmosphere.

  If you already have `build_romfs` in your PATH, then that will still
  take precedence.  (This seemed like the nicer option, though I don't
  have strong opinions.)


SciresM:
Atmosphere is a py2 project -- I don't actually have py3 installed myself, and greatly prefer py2. This is a point of dogma, sorry. Happy to accept a polyglot solution though, perhaps invoke python3 if python not detected?

The other stuff seems sensible enough.

comex:
huh, really.  I thought you just had `python` pointing to Python 3.  I will adjust.

comex:
Okay, updated:
- Changed the makefiles to only use `python3` if `python` doesn't exist
- Removed the shebang changes completely.
- Added another workaround (`MAKE`)

SciresM:
Again, much obliged ^^

PR #2617, opened 2025-09-04
build: enable parallel builds with automatic single-thread fallback

Lenflammeur:
### Summary
This change updates the top-level Makefile to:
- Detect available CPU cores with `nproc` and use them for parallel builds by default.
- Automatically retry a target with `-j1` if the parallel build fails (e.g. due to race conditions in certain modules).
- Apply the same logic to `clean` targets.

### Rationale
Most contributors benefit from faster builds when using all CPU cores.  
However, some targets (such as `haze`) can occasionally fail in parallel mode.  
This change improves build speed for the majority of cases while preserving reliability by falling back to single-thread mode automatically.

### Testing
Tested on Linux (WSL2) with devkitPro toolchain:
- `make` → builds in parallel successfully.
- Simulated parallel failure → retried automatically with `-j1` and completed.
- `make -j1` → still works as expected.
- `make clean` → cleans in parallel, falls back to `-j1` if needed.

No changes to target names, variables, or build outputs.  
Existing workflows remain fully compatible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,18 +4,19 @@ clean: clean-nx_release
 
 THIS_MAKEFILE     := $(abspath $(lastword $(MAKEFILE_LIST)))
 CURRENT_DIRECTORY := $(abspath $(dir $(THIS_MAKEFILE)))
+JOBS ?= $(shell nproc)
 
 define ATMOSPHERE_ADD_TARGET
 
 ATMOSPHERE_BUILD_CONFIGS += $(strip $1)
 
 $(strip $1):
-	@echo "Building $(strip $1)"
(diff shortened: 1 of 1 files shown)


SciresM:
This is somewhat incompatible with my personal workflow, which requires I pass make -jN where N is less than all my CPU cores, else my PC bluescreens. I am also not sure this behavior is preferable to just having you type make again on fail.

Though, even more ideally, the possibility of failure would be fixed? Please talk to me on discord if you particularly want to get this in. Otherwise, I will close in 24 hours.

SciresM:
Also worth noting: any merged build-system change must be tested on Windows, which is the only dev environment I use.

souldbminerr:
Atmosphere takes like 1 hr to build without bleeding cpu cores, with it it's like 10mins

SciresM:
Closing due to non-response.

PR #2638, opened 2025-11-15
Decrease ExtraSystemMemorySizeAtmosphere to 24_MB

hexbyt3:
Web Applet (LibAppletWe) is crashing with a memory error (0x800D2) in games like Scarlet/Violet.  Fixed similarly in commit 7a69f2f06 in 2023.


SciresM:
This is a bogus pull-request that does not fix the issue.

1) that value in pm-spec is only for firmware 1.0.0-4.0.0, after those versions dynamic determination is made using kernel-svc-info values.
2) Even when 0x800d2 is fixed, ro sysmodule has a further issue (0x3ec03), telling me you didn't actually test this. I am literally investigating this right now.

Please don't waste my time with bullshit pull requests in the future.

PR #2734, opened 2026-02-23
libvapours: add R_DISCARD macro

ndeadly:
This adds a new result macro `R_DISCARD` for explicitly discarding result values, avoiding the need to wrap expressions in `static_cast<void>(...)` when keeping the compiler happy after `ams::Result` was marked `[[nodiscard]]` in https://github.com/Atmosphere-NX/Atmosphere/commit/3bc19518202bdcfbda4caf2712e6a12f171ba06

### DIFF
--- a/libraries/libvapours/include/vapours/results/results_common.hpp
+++ b/libraries/libvapours/include/vapours/results/results_common.hpp
@@ -452,6 +452,9 @@ namespace ams::result::impl {
         }                                                                     \
     }
 
+/// Explicitly discards the result returned by an expression.
+#define R_DISCARD(res_expr) (static_cast<void>(res_expr))
+
 #if defined(ATMOSPHERE_BOARD_NINTENDO_NX) && defined(ATMOSPHERE_IS_STRATOSPHERE) && !defined(AMS_ENABLE_DETAILED_ASSERTIONS) && !defined(AMS_BUILD_FOR_DEBUGGING) && !defined(AMS_BUILD_FOR_AUDITING)
     #define AMS_CALL_ON_RESULT_ASSERTION_IMPL(cond, val) do { ::ams::diag::impl::FatalErrorByResultForNx(val); AMS_INFINITE_LOOP(); AMS_ASSUME(false); } while (false)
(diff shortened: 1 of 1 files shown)


ndeadly:
> Is it worth adding another commit to replace existing `static_cast<void>()` usages with this macro?

That did cross my mind, however I didn't want to waste my time tracking down every usage (not every `static_cast<void>()` applies to an expression returning `ams::Result`) without maintainer approval first.

The refactor to account for `[[nodiscard]]` is also incomplete, so maybe a second PR fixing all the compiler warnings and introducing `R_DISCARD` usage in one go might make more sense. I was going to attempt this, but then I saw the original intent was to match Nintendo's behaviour regarding how results are handled. I'm not sure I have the time/patience to go cross-checking all of that...

ndeadly:
Or maybe this should be `AMS_DISCARD` so that it's not specific to `ams::Result`? Or implement both, since it's nice to use `R_DISCARD` alongside other `R_`-prefixed macros.

Opinions welcome.

Ntemis:
@hexkyz can you review this please? You R the only one that can merge now

souldbminerr:
I second this. Compiling just ldr causes quite a lot of warnings under unused result, this macro could be wrapped around the problematic calls if checking the result is unsuitable

hexkyz:
This will be merged into master after 22.0.0 support is finished.

@ndeadly R_DISCARD sounds more accurate to me. Given that the ```[[nodiscard]]``` changes were restricted to nn::Result, it makes sense to have the macro be part of the result family. Unless there's an use case for discarding outside of ```ams::Result``` I'm not aware of?

souldbminerr:
Btw if this PR does get included some code should be changed to use this macro / check result, when I attempted to compile atmosphere there are numerous warnings for unused results.

ndeadly:
> This will be merged into master after 22.0.0 support is finished.
> 
> @ndeadly R_DISCARD sounds more accurate to me. Given that the ```[[nodiscard]]``` changes were restricted to nn::Result, it makes sense to have the macro be part of the result family. Unless there's an use case for discarding outside of ```ams::Result``` I'm not aware of?

The main use case is for ams::Result, so I think R_DISCARD fits best. No other specific use case, I was just reasoning that explicitly ignoring the return value of a function was not necessarily unique to ams::Result

PR #2711, opened 2026-01-07
add exosphere config to control memory mode

lulle2007200:

PR #2748, opened 2026-03-19
[Feature] Auto dram cfg & Auto memory mode

CTCaer:
This PR adds the following:

`exosphere: automatically adjust dram id if needed`:
Support for automatic detection and adjusting of DRAM ID reporting when unit has an 8GB RAM mod and fuses couldn't be burnt.

By comparing the actual physical size from the locked MC registers and DRAM ID supposed size, the id gets adjusted when it doesn't match.
This allows PCV to correctly train the ram frequencies needed for avoiding crashes or degraded performance.

`exosphere: allow memory mode to be used on retail`
A simple flag configured at bootloader time allows user to bypass the 4GB max ram allocation limit on retail units by enabling auto memory mode.
Enabling this with 8GB ram, sets max to 8GB, otherwise it gets limited to stock max 4GB.
On development units or devices with 4GB, it does nothing since ram config matches max size limit anyway.

By previous communication with @lulle2007200, the second part of this PR replaces https://github.com/Atmosphere-NX/Atmosphere/pull/2711
The new implementation is compacted to actually enable NX boot config for retail units.
The default at this time is auto memory mode which should cover 100% of actual real use cases, but still allows additions in the future.

To sum up:
1st part is for allowing automatic dram config detection with 8GB ram mods.
2nd part is for allowing user to choose between 4GB or 8GB max ram limit in such cases.

EDIT:
The automatic detection also allows T210 units with 8GB ram mods and burnt fuse dram id to 7, to boot too.

### DIFF
--- a/config_templates/exosphere.ini
+++ b/config_templates/exosphere.ini
@@ -15,6 +15,13 @@
 #  Desc: Controls whether userland has access to the PMU registers.
 #  NOTE: It is unknown what effects this has on official code.
 
+# Key: enable_mem_mode, default: 0.
+#  Desc: Controls whether boot config memory mode is taken into account
+#   for retail units. This does not affect development units.
+#  NOTE: On retail units max ram size is capped to 4GB.
+#   Enabling this will use the boot config memory mode parameter,
+#   which by default is auto and size gets set based on physical size.
+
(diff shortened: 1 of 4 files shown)


hexkyz:
Approved for merging into master after 22.0.0 support.

PR #2750, opened 2026-03-20
ams: add enum recognition for 19.0.2

CosmicCurator:
Also recognize the out of step Chinese exclusive 19.0.2 firmware release.

### DIFF
--- a/libraries/libstratosphere/source/fs/impl/fs_id_string_impl.os.generic.cpp
+++ b/libraries/libstratosphere/source/fs/impl/fs_id_string_impl.os.generic.cpp
@@ -42,7 +42,7 @@ namespace ams::fs::impl {
             case KeyGeneration_16_0_0: return "16.0.0-16.0.3";
             case KeyGeneration_17_0_0: return "17.0.0-17.0.1";
             case KeyGeneration_18_0_0: return "18.0.0-18.1.0";
-            case KeyGeneration_19_0_0: return "19.0.0-19.0.1";
+            case KeyGeneration_19_0_0: return "19.0.0-19.0.2";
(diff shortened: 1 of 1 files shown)


alula:
This is likely unnecessary because `19.0.2` identifies itself as `19.0.1` internally, just has a different display version

CosmicCurator:
Unnecessary in what respect? It is unnecessary from a functionality perspective since Atmosphere will always attempt to run micro version increments and can run 19.0.2 without any issues. 

Other than that Atmosphere recognizes all released firmware versions internally and 19.0.2 is a unique firmware version with multiple changes, I added this primarily for posterity. In terms of identification I can only assume you are referring to the unchanged `NintendoSDK Firmware for NX 19.0.1-1.0` full version string?

hexkyz:
This is indeed unnecessary. It's not exactly true that Atmosphère recognizes all released firmwares, at least not in the way you expect.
My reasoning here is:
- The SystemVersion program still refers to it as "NintendoSDK Firmware for NX 19.0.1-1.0" and shares the same digest with 19.0.1. It does have a different version hash but we've never accounted for that in Atmosphère. A good example of identical version numbers with different version hashes are production vs. development firmwares, which we don't support.
- "19.0.2" is only recognized as a program version (1275199548);
- If we were to follow the same logic, program versions 0.0.0 (450) and 0.0.1 (65796), for example, would need to be included as well. These were the actual program versions back in the 1.0.0 and 2.0.0 firmware era.

I agree that 19.0.2 is a bit of an odd case and an argument could be made on the grounds that it had an actual, distinguishable release. However, given the ambiguity, the more technically accurate version would be the SystemVersion program's contents and, therefore, we should keep treating 19.0.2 as 19.0.1.

PR #2763, opened 2026-04-01
Fix off by 1 bug + allocation error + missing initialize in PGL

masagrator:
Off by 1 bug:

For reference we need to know this variable:
https://github.com/Atmosphere-NX/Atmosphere/blob/61ac03e22d20460f2032a2b733d35a86bbc870e0/libraries/libstratosphere/include/stratosphere/fs/fs_mount.hpp#L22

Now in source file we can find that mount name has exactly 16 bytes (15 + null)
```cpp
        constexpr inline char HostPackageMountName[] = "HostPackageRead";
        static_assert(sizeof(HostPackageMountName) - 1 <= fs::MountNameLengthMax);
```
15 <= 15
It's OK.

m_mount_name is declared to have 15 bytes
```
char m_mount_name[fs::MountNameLengthMax]   = {};
```

so that check:

```cpp
R_UNLESS(strlen(mount) <= sizeof(m_mount_name) - 1, pgl::ResultBufferNotEnough());
```

originally results in:
```
15 <= 14
```
So it always returns error, the rest of function is removed by compiler and replaced with inlined deconstructor.

This change follows how other places in code declare variables with `fs::MountNameLengthMax` as size.

https://github.com/Atmosphere-NX/Atmosphere/blob/61ac03e22d20460f2032a2b733d35a86bbc870e0/libraries/libstratosphere/source/fs/fsa/fs_mount_name.hpp#L22

Sorry for a little mess in commits. I was changing the actual fix to not break other things. You should squash that. 

---

Edit: Also PGL uses `ams::ncm::ReadContentMetaPathWithoutExtendedDataOrDigest()` function that is using `operator new[]`, but because it's not overriden it defaults to stubbed malloc which in result returns AllocationFailed error in functions that use it. FIxed that by overriding new and delete operators.

---

Edit2:
Also found that this line:
https://github.com/Atmosphere-NX/Atmosphere/blob/61ac03e22d20460f2032a2b733d35a86bbc870e0/libraries/libstratosphere/source/pgl/srv/pgl_srv_shell_host_utils.cpp#L322

to work properly requires `ams::lr::g_location_resolver_manager` to be initialized, but it's not, so sysmodule will segfault when invoking `OpenLocationResolver()`. Added fix for that.

### DIFF
--- a/stratosphere/pgl/source/pgl_main.cpp
+++ b/stratosphere/pgl/source/pgl_main.cpp
@@ -37,6 +37,7 @@ namespace ams {
             R_ABORT_UNLESS(pmshellInitialize());
             R_ABORT_UNLESS(ldrShellInitialize());
             R_ABORT_UNLESS(lrInitialize());
+            lr::Initialize();
 
             /* Verify that we can sanely execute. */
             ams::CheckApiVersion();


alula:
`git rebase -i 22_support` -> squash commits

`git push --force` and probably target this towards 22_support branch

masagrator:
It can be also squashed with "Squash and merge" button 👍

Also i don't consider this part of 22.0.0 support. If repository owner considers this a necessity to accept PR I will make another PR.

PR #2770, opened 2026-04-04
Discard unused ams::Result in libstratosphere with R_DISCARD macro

masagrator:
Suppresses all warnings about nodiscard.

### DIFF
--- a/libraries/libstratosphere/source/ncm/ncm_install_task_base.cpp
+++ b/libraries/libstratosphere/source/ncm/ncm_install_task_base.cpp
@@ -807,7 +807,7 @@ namespace ams::ncm {
 
     Result InstallTaskBase::PrepareSystemUpdateDependency() {
         /* Cleanup on failure. */
-        ON_RESULT_FAILURE { this->Cleanup(); };
+        ON_RESULT_FAILURE { R_DISCARD(this->Cleanup()); };
 
         /* Count the number of content meta entries. */
         s32 count;


ndeadly:
Did you check these cases against Nintendo's behaviour where applicable? That was the reason Scires didn't deal with the warnings himself https://github.com/Atmosphere-NX/Atmosphere/commit/3bc19518202bdcfbda4caf2712e6a12f171ba064

> All warnings should be fixed to do the same thing Nintendo does as relevant, but this
> is taking a phenomenally long time and is not actually the most important work to do,
> so it can be put off for some time to prioritize other tasks for 21.0.0 support.

Looking at the way the rest of the code is written, it seems like good form to add a comment to explain that Nintendo behaviour is being matched before discarding the result of a call. In other cases the unhandled result may not have been intentional. E.g. I'm not sure we should be ignoring the result of `os::CreateThread`, `AMS_ABORT_UNLESS` probably makes more sense in this case https://github.com/masagrator/Atmosphere/blob/e7faea2b16c1498f4bad3c664c29f17984040506/libraries/libvapours/source/sdmmc/impl/sdmmc_device_detector.cpp#L203

masagrator:
I didn't. 

I'm just bringing back old behavior since most of compilation time in case of my sys-ticon is printing errors related to nodiscard in libstratosphere headers, i edited cpp additionally. This won't change how current implementation works.

If someone is already taking task of checking which portions of code need different approach or they are planning to do it in foreseeable future, i can close this. But I don't see the reason to not do it when noone is willing to take the task and hinder developers experience because of pure design choices. 

SciresM is not working on this anymore so who else?

Also Nintendo doesn't always takes the most sensible approach looking at what was commented already as not checked. 😅

ndeadly:
Yeah, the error messages are certainly annoying, but presumably the point in marking `ams::Result` `[[nodiscard]]` in the first place was to bring attention to the unhandled results in the codebase so they can be considered on an individual basis and addressed appropriately. If we are just going to blindly discard the results to suppress the warnings we might as well just remove the `[[nodiscard]]` qualifier

PR #2786, opened 2026-04-07
ams: add enum recognition for 22.1.0

alula:


spydersmash:
I was so confused I thought I broke my switch but apparently I accidentally updated to 22.1.0 instead of just 22.0.0 💀. 

Hope this is updated soon! Thank you guys for the work you do

PR #2806, opened 2026-05-12
Security fixes within stratosphere

Silverdragon122:
Summary

Harden PGL IPC string handling and fix HTCFS rename request lengths.

Changes

Copy PGL IPC path and argument buffers into bounded, NUL-terminated local strings before passing them to existing C-string helpers.

Reject oversized snapshot dumper arguments when formatting would truncate the local argument buffer.

Fix HTCFS RenameFile and RenameDirectory request construction so old_path_len is based on old_path and new_path_len is based on new_path.

Impact

Malformed local inputs could previously cause out-of-bounds reads, malformed requests, service crashes, or limited adjacent memory disclosure. This requires local service access or HTCFS availability.

Compatibility

No public API or IPC ABI changes. Valid callers continue to use the same inputs. Invalid overlong or unterminated inputs now fail cleanly instead of relying on adjacent memory.

### DIFF
--- a/libraries/libstratosphere/source/htcfs/htcfs_client_impl.cpp
+++ b/libraries/libstratosphere/source/htcfs/htcfs_client_impl.cpp
@@ -532,8 +532,8 @@ namespace ams::htcfs {
         Header request, response;
 
         /* Create header for the request. */
-        const auto old_path_len = std::strlen(new_path);
-        const auto new_path_len = std::strlen(old_path);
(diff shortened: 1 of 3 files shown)


hexkyz:
This is very obviously LLM generated. Closed.

bakugo:
Isn't the `old_path_len`/`new_path_len` thing an actual bug though?

PR #2821, opened 2026-07-14
sf: close request copy handles in mitm ForwardRequest

dogty:
ForwardRequest closes copy handles from the forward response but never from the request. Any mitm'd command that carries a client copy handle therefore leaks that handle in the mitm process for its lifetime. A leaked transfer memory pins the client process object after it dies, its application memory reservation never releases, and the next application launch blocks forever freezing the console.

Close the request's copy handles after the forward send, parsed from the saved message.


hexkyz:
Looks fine, but this cleanup should be in an ON_SCOPE_EXIT so that it runs immediately after the R_TRY svc.

dogty:
Thank you for the feedback wasn't aware of this macro. Fixed it.

PR #2832, opened 2026-08-31
Upgrade Scripts for Python 3.x

impeeza:
Upgrade of the scripts 
* utilities\erpt.py
* utilities\nxo64.py
* fusee\program\update_mtc_tables.py
* fusee\program\update_sdram_params.py

to replace the deprecated and removed function `xrange` for the function `range` which is on Python 3 and onward

### DIFF
--- a/fusee/program/update_mtc_tables.py
+++ b/fusee/program/update_mtc_tables.py
@@ -43,10 +43,10 @@ def main(argc, argv):
             assert board.startswith('T210Sdev')
             soc = 'erista'
             count = 10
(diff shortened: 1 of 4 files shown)


SciresM:
I may be retired/not working on it, but atmosphere is a py2 project. If you really want this in, talk to hexkyz first. It is important that you test with python2 to make sure py2 doesn't break if you're going for cross-compat.

ELY3M:
I know python2 is EOL now.   I think it is best to add both functions for both python2 and python3    add in a check to check which python you are using and use that.   like this   if (python2)  use python2 else use python3

masagrator:
...that's what SciresM said.

PR #2835, opened 2026-09-04
RO exception offsets are signed

masagrator:
Assassin's Crees 3 NROs have .eh_frame_hdr above MOD0 which proves that those are signed. I couldn't load NRO to IDA with TSRBerry's nxo64 which also uses unsigned offsets so i sent PR to TSRBerry fork that fixes this issue and NRO loads correctly.


hexkyz:
Hm... I'll need to look a bit into this. Reversing the relevant stuff that deals with NROs shows me that the exception offsets were and still are u32s.

